PR #11, opened 2015-12-31
fix(.travis.yml,_scripts): add to CI

### LINKED ISSUE #9: setup CI
This issue requires #5. For this issue to be accepted, CI should run `make docker-build` on all branches, on every commit.

Then, on pushes to master, CI should also run `make docker-push` to push our e2e testing image to quay and dockerhub. This functionality implies that we'll need a deploy script like [the one in deis/builder](https://github.com/deis/builder/blob/master/_scripts/deploy.sh)

arschles:
Fixes #9


_scripts/deploy.sh
```diff
@@ -5,7 +5,7 @@
 
 cd "$(dirname "$0")" || exit 1
 
-export IMAGE_PREFIX=deis VERSION=2.0.0-alpha
+export IMAGE_PREFIX=deis VERSION=canary
```
arschles: @sgoings @jchauncey thoughts on pushing `canary` on each build?

sgoings: I like that idea. I'll probably be using this convention across all Deis components if we agree this is where we want to go re: HEAD version.

arschles: :+1: 

cc/ @technosophos and @mboersma

technosophos: I like the idea, too. I don't have really strong opinions about the name, but I admit that I think canary is sorta cute. :bird:

technosophos:
Most of this is actually just removing lots of dead code left over from deis/workflow, right?

arschles:
yup

PR #21, opened 2015-12-31
fix(tests): change package names

arschles:
so that golint doesn’t complain about underscores in package names

### DIFF
--- a/tests/apps_test.go
+++ b/tests/apps_test.go
@@ -1,4 +1,4 @@
-package _tests_test
+package tests
 
 import (
 	. "github.com/onsi/ginkgo"
--- a/tests/auth_test.go
+++ b/tests/auth_test.go
@@ -1,4 +1,4 @@
-package _tests_test
+package tests
 
 import (
 	. "github.com/onsi/ginkgo"
--- a/tests/builds_test.go
+++ b/tests/builds_test.go
@@ -1,4 +1,4 @@
-package _tests_test
+package tests
 
 import (
 	. "github.com/onsi/ginkgo"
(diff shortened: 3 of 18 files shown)


technosophos:
Must appease the `go lint`. Rename your packages and sacrifice a goat.

arschles:
Hey, this PR gives us 100% across the board, *in addition* to the A+ on http://goreportcard.com/report/deis/workflow-e2e. Perfection!

carmstrong:
> this PR gives us 100% across the board

:+1:

technosophos:
Well in that case, you can keep your goat. :goat:

sgoings:
Except Travis hates it. :stuck_out_tongue_winking_eye: https://travis-ci.org/deis/workflow-e2e/builds/99633209

arschles:
@sgoings yea, needs moar https://github.com/deis/workflow-e2e/pull/11 first, then travis will :heart: it

technosophos:
@sgoings Travis will continue to hate until #11 gets two LGTMs.

mboersma:
I had assumed that test code didn't need linting or format checking, but this is better. 

As background, mostly to remind myself: Ginkgo assumes there is "real" go code in the package to test, and generates a new package name with the `_test` suffix (which go allows to coexist, apparently). Then the tests run the package from the outside for more proper "black box" testing. When we write unit tests for other go code using ginkgo, this is a good feature.

But in this case there is no such "real" go code to test: the e2e suite stands on its own. And that generated package name was ugly. :+1:

arschles:
>Ginkgo assumes there is "real" go code in the package to test, and generates a new package name with the _test suffix (which go allows to coexist, apparently)

@mboersma I was wondering why those package names, but makes perfect sense now - Ginkgo's trying to be a good citizen. Thanks for explaining

PR #24, opened 2016-01-12
chore(tests): update for new Django REST framework

mboersma:
This fixes several tests which broke when DRF was updated in https://github.com/deis/workflow/commit/90b02356f9fe7cb49f6f6300892d61edf4cf6fbc.


carmstrong:
This also makes the tests more polite by shouting less. :+1:

helgi:
I didn't even keep in the new default pagination DRF has! :D

PR #25, opened 2016-01-13
fix(deploy.sh): push docker images to deisci/ orgs

mboersma:
Changes the repository org to `deisci` for consistency with the other images that are pushed for Deis v2.

### DIFF
--- a/_scripts/deploy.sh
+++ b/_scripts/deploy.sh
@@ -5,7 +5,7 @@
 
 cd "$(dirname "$0")" || exit 1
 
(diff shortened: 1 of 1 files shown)


arschles:
@mboersma what do you think about releasing an official beta of these tests when we release beta?

mboersma:
> an official beta of these tests

Why not? We should tag the `deis-e2e:alpha-v2` test image as well and release the test image in sync with all the other components IMHO.

arschles:
@mboersma :+1:

PR #27, opened 2016-01-27
fix(tests): clean up .git and fix config:list syntax

mboersma:
Fixes two broken test cases in apps_test.go.

### DIFF
--- a/tests/apps_test.go
+++ b/tests/apps_test.go
@@ -1,11 +1,12 @@
 package tests
 
 import (
+	"time"
+
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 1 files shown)


arschles:
are these tests running in CI at all yet?

mboersma:
> are these tests running in CI at all yet?

Not yet; getting there. I do have these tests running as a k8s [job](http://kubernetes.io/v1.1/docs/user-guide/jobs.html) though and I'm going to PR that today.

arschles:
>Not yet; getting there

Cool!

>I do have these tests running as a k8s job though and I'm going to PR that today.

Great to hear. I'm looking forward to having these running in CI

PR #28, opened 2016-01-29
ref(tests): dot-import two packages for readability

mboersma:
Generally I'm not a fan of Go dot-imports, but for this test code I think it makes it more approachable and easier to read.


arschles:
fair nuff. LGTM

PR #33, opened 2016-02-09
feat(apps): verify that a bogus app url returns a 404

vdice:


sgoings:
Since this has 2 LGTMs this is good to merge now.

PR #34, opened 2016-02-11
fix(*): prevent auth:cancels from colliding

sgoings:


sgoings:
Pushed a few changes that clean up code. We had two incantations of login and cancel, and the [login,cancel]Sess version made it possible to fall into a race condition due to the use of an external deis cli. Since we were just using [login,cancel]Sess, I swapped usage to just use [login,cancel].

tests/tests_suite_test.go
```diff
@@ -121,14 +121,10 @@ var _ = AfterEach(func() {
 })
 
 var _ = AfterSuite(func() {
-	cancelUserSess, cancelUserErr := cancelSess(url, testUser, testPassword)
-	cancelAdminSess, cancelAdminErr := cancelSess(url, testAdminUser, testAdminPassword)
+	os.Chdir(testHome)
```
sgoings: This is necessary to avoid the madness of not having a directory structure to work in:

```
Logged in as test-81
•SSSSSSSSSSSSSSSSSSSSSSSSSshell-init: error retrieving current directory: getcwd: cannot access parent directories: No such file or directory
Logged in as test-81
shell-init: error retrieving current directory: getcwd: cannot access parent directories: No such file or directory
Please log in again in order to cancel this account
Logged in as test-81
Account cancelled
shell-init: error retrieving current directory: getcwd: cannot access parent directories: No such file or directory
Logged in as admin
shell-init: error retrieving current directory: getcwd: cannot access parent directories: No such file or directory
Please log in again in order to cancel this account
Logged in as admin
Account cancelled
```

mboersma:
LGTM++. Also, how do you pronounce `Sess`? :smiling_imp:

sgoings:
I am on a mission to remove all things in the world that sound like Seth. First victory: Sess.

PR #35, opened 2016-02-11
feat(apps): verify that the app url can be opened

vdice:
Test still X'ed out, but can verify that it fails properly 
(since implementation under test is broken at time of writing) 
by removing the 'X' in front of the 'It("can open the app's URL"...'


shims/system.go
```diff
@@ -0,0 +1,67 @@
...
+	OutFile  *os.File
+	ShimFile *os.File
+}
+
+func CreateSystemShim(toShim string) Shim {
+	tempDir := strings.TrimSuffix(os.TempDir(), "/")
+	// create out file for shim to write to
+	outFile, err := ioutil.TempFile(os.TempDir(), fmt.Sprintf("%s.out", toShim))
+	if err != nil {
+		panic(err)
```
vdice: What is our standard error handling practice?

mboersma: Generally you return an `error` (that may be `nil`) from every function and force its caller to handle it. `panic()` is mostly frowned upon since it makes the control flow hidden. Although test code is one of the few areas where `panic` may be justifiable.

So you could change this signature to return `(Shim, error)` and then handle it in the caller. Or just leave it as-is in this case since not being able to get a temp file is a rare and more or less catastrophic occurrence that might justify a `panic`.

PR #36, opened 2016-02-11
fix(keys): list keys, create + remove temporary key

sgoings:


sgoings:
There were two things wrong with the original incantation:

- I reused the variable: randSuffix which was an idiotic thing to do... so I was still deleting the user's only ssh key
- I never added the user's new ssh key to deis, I only created the key and then tried to have deis remove it.

PR #40, opened 2016-02-18
feat(apps): enable deis open test

vdice:


Makefile
```diff
@@ -3,7 +3,7 @@ export GO15VENDOREXPERIMENT=1
 SHORT_NAME := deis-e2e
 
 SRC_PATH := /go/src/github.com/deis/workflow-e2e
-DEV_IMG := quay.io/deis/go-dev:0.5.0
+DEV_IMG := quay.io/deis/go-dev:0.7.0
```
sgoings: latest is 0.9.0 https://github.com/deis/docker-go-dev/releases

arschles: note that 0.8.0 gives the updated glide version, 0.9.0 gives you that and go 1.6. they're intentionally separate tags

mboersma:
See also #41.

PR #42, opened 2016-02-19
fix(apps_test): Fix the error message we check for

jchauncey:


tests/apps_test.go
```diff
@@ -36,7 +36,7 @@ var _ = Describe("Apps", func() {
 			sess, err := start("deis logs -a %s", appName)
 			Expect(err).To(BeNil())
 			Eventually(sess).Should(Exit(1))
-			Eventually(sess.Err).Should(Say("Not found."))
+			Eventually(sess.Err).Should(Say(`Error: There are currently no log messages. Please check the following things:`))
```
sgoings: What behavior do the tickmarks give?

jchauncey: Normally you use them when doing multiline strings. I was attempting to do that but decided against it. So I forgot to remove them here. The gomega docs actually them them instead of double quotes - 

```
    It("should stream data", func() {
        Eventually(buffer).Should(gbytes.Say(`Attached to stream as client \d+`))

        client.ReticulateSplines()
        Eventually(buffer).Should(gbytes.Say(`reticulating splines`))
        client.EncabulateRetros(7)
        Eventually(buffer).Should(gbytes.Say(`encabulating 7 retros`))
    })
```

jchauncey: if you want i can remove them.

sgoings: Oh I just didn't know what they did for us, that's all. Not holding this up. Just trying to learn.

PR #49, opened 2016-02-21
fix(cancel): delete apps before auth:cancel

mboersma:
Avoids a 409 error and improves repeatability of the suite. See deis/workflow#391.

### DIFF
--- a/tests/tests_suite_test.go
+++ b/tests/tests_suite_test.go
@@ -168,8 +168,20 @@ func cancel(url, username, password string) {
 	// log in to the account
 	login(url, username, password)
 
+	// remove any existing test-* apps
+	sess, err := start("deis apps")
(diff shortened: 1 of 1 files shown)


tests/tests_suite_test.go
```diff
@@ -168,8 +168,20 @@ func cancel(url, username, password string) {
 	// log in to the account
 	login(url, username, password)
 
+	// remove any existing test-* apps
```
vdice: Only suggestion is if there are other places besides `cancel` that might want to remove the test apps in such fashion, then maybe we can extract this logic into its own helper func, `destroy` or `destroyAllApps`.

mboersma: That's a good point--I should have done it that way. We can easily extract it into a different `func` if need be in the future.

PR #51, opened 2016-02-21
fix(apps_test): update empty apps:logs expectations

mboersma:
A test case was failing because now even when the logging subsystem is properly configured, if there are no app logs the following is emitted on `stdout`:
```console
$ deis logs -a appname
There are currently no log messages. Please check the following things:
1) Logger and fluentd pods are running.
2) If you just installed the logger components via the chart, please make sure you restarted the workflow pod.
3) The application is writing logs to the logger component.
You can verify that logs are appearing in the logger component by issuing the following command:
curl http://<log service ip>:8088/logs/test-824225533 on a kubernetes host.
To get the service ip you can do the following: kubectl get svc deis-logger --namespace=deis
```
This updates the test case to match the current behavior so it passes. Or do we consider this new behavior a bug? cc: @krancour @jchauncey


jchauncey:
How is logging running on gke? Daemonsets aren't enabled by default so are
we doing something to turn them on? Also workflow has to be restarted after
the logging service starts.

mboersma:
Never mind--I hadn't updated my local `deis` CLI for testing, so this test actually passes currently without this PR.

PR #53, opened 2016-02-21
feat(ps_test): implement table-driven processes tests

### LINKED ISSUE #17: write the ps tests
they're all commented out: https://github.com/deis/workflow-e2e/blob/master/tests/ps_test.go

mboersma:
Uses ginkgo's [table-driven-tests](http://onsi.github.io/ginkgo/#extensions) extension to test likely `ps:scale` and `ps:restart` scenarios.

Closes #17.

TODO:
- [x] fix `Should(Say("done"))` sections that are commented out
- [x] ~~figure out dynamically whether it's a `cmd` or `web` proc type~~ I think there's a better solution that can build off #57
- [x] some problem with escaping the `curl` command line?

### DIFF
--- a/tests/ps_test.go
+++ b/tests/ps_test.go
@@ -1,44 +1,155 @@
 package tests
 
 import (
+	"fmt"
+	"math/rand"
+	"regexp"
+	"sort"
+	"strconv"
+	"sync"
+
 	. "github.com/onsi/ginkgo"
-	// . "github.com/onsi/gomega"
+	. "github.com/onsi/ginkgo/extensions/table"
+	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gbytes"
+	. "github.com/onsi/gomega/gexec"
 )
 
-var _ = Describe("deis", func() {
+var procsRegexp = `(%s-v\d+-[\w-]+) up \(v\d+\)`
+
+// scrapeProcs returns the sorted process names for an app from the given output.
+// It matches the current "deis ps" output for a healthy container:
+//   earthy-vocalist-v2-cmd-1d73e up (v2)
+//   myapp-v16-web-bujlq up (v16)
+func scrapeProcs(app string, output []byte) []string {
+	re := regexp.MustCompile(fmt.Sprintf(procsRegexp, app))
+	found := re.FindAllSubmatch(output, -1)
+	procs := make([]string, len(found))
+	for i := range found {
+		procs[i] = string(found[i][1])
+	}
+	sort.Strings(procs)
+	return procs
+}
+
+var _ = Describe("Processes", func() {
+
 	Context("with a deployed app", func() {
 
-		XIt("can scale upward", func() {
-			// "deis ps:scale web=5 --app=%s"
-			// "deis ps:list --app=%s"
-			// curl app
-			// "deis ps:scale web=1 --app=%s"
-			// "deis ps:list --app=%s"
-			// curl app
+		var testApp App
+		once := &sync.Once{}
+
+		BeforeEach(func() {
+			// Set up the Processes test app only once and assume the suite will clean up.
+			once.Do(func() {
+				testApp = deployApp("example-go")
+			})
 		})
 
-		XIt("can scale down to 0", func() {
-			// "deis ps:scale web=0 --app=%s"
-			// "deis ps:list --app=%s"
-			// curl app
-			// "deis ps:scale web=1--app=%s"
-			// "deis ps:list --app=%s"
-			// curl app
-		})
+		DescribeTable("can scale up and down",
 
-		XIt("can restart all processes", func() {
-			// "deis ps:scale web=5 --app=%s"
-			// "deis ps:list --app=%s"
-			// curl app
-			// "deis ps:restart web --app=%s"
-			// "deis ps:list --app=%s"
-			// curl app
-		})
+			func(scaleTo, respCode int) {
+				// TODO: need some way to choose between "web" and "cmd" here!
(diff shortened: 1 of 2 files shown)


tests/ps_test.go
```diff
@@ -1,44 +1,156 @@
...
+
+		DescribeTable("can restart processes",
+
+			func(restart string, scaleTo int, success bool, respCode int) {
+				// TODO: need some way to choose between "web" and "cmd" here!
+				// scale the app's processes to the desired number
+				sess, err := start("deis ps:scale web=%d --app=%s", scaleTo, appName)
+				Expect(err).NotTo(HaveOccurred())
+				Eventually(sess).Should(Say("Scaling processes... but first,"))
+				// Eventually(sess).Should(Say(`done in \d+s`))
```
vdice: Can these be removed?

tests/ps_test.go
```diff
@@ -1,44 +1,156 @@
...
+				Eventually(sess).Should(Say(`done in \d+s`))
+				Eventually(sess).Should(Say("=== %s Processes", appName))
+				Eventually(sess).Should(Exit(0))
+
+				// test that there are the right number of processes listed
+				sess, err = start("deis ps:list --app=%s", appName)
+				Expect(err).NotTo(HaveOccurred())
+				Eventually(sess).Should(Say("=== %s Processes", appName))
+				Eventually(sess).Should(Exit(0))
+				re := regexp.MustCompile(fmt.Sprintf(`%s-v\d-[\w-]+ up \(v\d\)`, appName))
```
vdice: What is the difference between this `re` and the one in `scrapeProcs` above?  Any chance they can be merged into one standalone, reusable regexp?

tests/ps_test.go
```diff
@@ -1,44 +1,156 @@
...
+				case "by wrong type":
+					// TODO: need some way to choose between "web" and "cmd" here!
+					arg = "cmd"
+				case "one":
+					arg = beforeProcs[rand.Intn(len(beforeProcs))]
+				}
+				sess, err = start("deis ps:restart %s --app=%s", arg, appName)
+				Expect(err).NotTo(HaveOccurred())
+				Eventually(sess).Should(Say("Restarting processes... but first,"))
+				// Eventually(sess).Should(Say(`done in \d+s`))
```
vdice: ditto as above -- can these be removed or do we need to uncomment and verify?

vdice:
Closes #17 (right @mboersma @arschles  ?)

mboersma:
These tests seem to point out a couple of issues in current Deis:
- sometimes after a `ps:restart` not all processes are listed right away
- after a `ps:scale` from 0 to 1 we get a 502 right away from the app, not a 200

Each of those is a potential showstopper that these tests uncovered. I will create issues for them after I reproduce them. 

Should these tests get merged before fixes are available? I'm curious what others think...

mboersma:
```console
$ DEIS_ROUTER_SERVICE_HOST=192.168.64.2 DEIS_ROUTER_SERVICE_PORT=31938 make test-integration
go test ./tests/... -v -ginkgo.slowSpecThreshold=120.0  # -ginkgo.v
=== RUN   TestTests
Running Suite: Deis Workflow
============================
Random Seed: 1456198747
Will run 17 of 70 specs

SSSSSSSSSSS•••••••••••••••••SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS
Ran 17 of 70 Specs in 272.901 seconds
SUCCESS! -- 17 Passed | 0 Failed | 0 Pending | 53 Skipped PASS | FOCUSED
```

vdice:
Awesome work, @mboersma !

tests/ps_test.go
```diff
@@ -1,44 +1,167 @@
...
-			// "deis ps:scale web=0 --app=%s"
-			// "deis ps:list --app=%s"
-			// curl app
-			// "deis ps:scale web=1--app=%s"
-			// "deis ps:list --app=%s"
-			// curl app
+		BeforeEach(func() {
+			// Set up the Processes test app only once and assume the suite will clean up.
+			once.Do(func() {
+				os.Chdir("example-go")
```
vdice: In #57, I created a helper func, `deployApp` that runs these steps.  See what you think of that please supply any suggestions/comments... we can potentially revisit this block and invoke that here, perhaps in 57 if this is merged by then.

mboersma: `deployApp` in #57 is great. I have some comments I'll leave over there. When that gets merged this should definitely be refactored to use that.

mboersma:
I'm going to refactor this to use `deployApp` from #57 once it's merged, then I'll ask for another quick review.

vdice:
@mboersma #57 has been merged; could you also knock out the TODO [here](https://github.com/deis/workflow-e2e/blob/master/tests/apps_test.go#L142) ?

mboersma:
This now uses `deployApp` and `procsRegexp` to be more [DettRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). If someone else could give it a review after the minor refactoring, then I think it's ready to go.

Although note that this seems to expose a few platform bugs that aren't always deterministic. For example, sometimes scaling from 0 to 1 yields a 404 if you hit the app endpoint right away as these tests do. 

But I think the test methodology is sound, so I'm of the opinion that it's better to merge valid tests that may fail and be forced to fix the code under test. I am curious to hear others' opinions.

mboersma:
Talking with @sgoings and @vdice about this, consensus is we should check in valid tests even if they go red, so I'm merging this.

PR #55, opened 2016-02-22
feat(apps_test): add further apps:info expectations

### LINKED ISSUE #37: apps_test: actually inspect 'app info' output
Currently the 'can get app info' test simply verifies that executing `deis info` does not produce an error.

It would be preferable to actually check the output values, e.g. those of `url`, `owner`, `id` among others.

vdice:
Closes #37


vdice:
The only thing we're not checking now are the `created` and `updated` items, though perhaps its easy enough to add regex checks for those...

PR #56, opened 2016-02-22
feat(deploy.sh): remove hardcoded VERSION to use git-<sha> declared in Makefile

vdice:
As part of the effort towards the promotion of immutable component images...


sgoings:
I'll have a corresponding PR to make sure we're using the most up to date workflow-e2e artifact in the deis-v2-e2e test job.

Issue #16, opened 2015-12-31
finish the perms tests


arschles:
there's one last `XIt` to finis: https://github.com/deis/workflow-e2e/blob/master/tests/perms_test.go#L63

vdice:
Closed via #57

PR #57, opened 2016-02-22
feat(perms_test): add perms:* invocations on deployed app

vdice:
Also created reusable `deployApp` func to be used among multiple tests

### DIFF
--- a/tests/tests_suite_test.go
+++ b/tests/tests_suite_test.go
@@ -25,6 +25,11 @@ type Cmd struct {
 	CommandLineString string
 }
 
+type App struct {
+	Name string
+	URL  string
+}
+
 const (
 	deisRouterServiceHost = "DEIS_ROUTER_SERVICE_HOST"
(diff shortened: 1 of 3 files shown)


vdice:
Closes #16

tests/tests_suite_test.go
```diff
@@ -25,6 +25,11 @@ type Cmd struct {
 	CommandLineString string
 }
 
+type App struct {
```
vdice: What do we think of using an extensible `App` struct for `testApp` instances, as opposed to the currently used/not quite as powerful `appName` string?

mboersma: This is the right solution--we need to keep URL separate from name. (We may also want to store whether it's a buildpack or Dockerfile or `deis pull` app in the `App` struct at some point, so we can vary the `deployApp` strategy and know whether to use `cmd` or `web`--see code `TODO:`s in #53).

PR #60, opened 2016-02-26
feat(reporters): add junit reporter if JUNIT=true

sgoings:


sgoings:
I have 2 LGTMs here, but I actually have a question...

Question would be: does it actually make more sense to have the env var be something like `JUNIT` or `JUNIT_XML`?

helgi:
I'd go `JUNIT` if you just going for that format, if you are planning on supporting many diff format via that as a switch statement, stick with `CI`

sgoings:
This has now been updated to use JUNIT instead of CI, @helgi. I've cleared the LGTMs so that people can re-review.

Issue #32, opened 2016-02-04
write tests for certificates


helgi:

vdice:
Will we want to dynamically generate (and remove) self-signed certs during test runtime via `openssl` or create a long-lived/scm'd `*.crt` and `*.key` file that these tests use?

helgi:
For `workflow` I went with pre-generated certs for simplicity and speed

helgi:
You could just copy those over, there are 3 sets of certs, one is wildcard with a SAN set

PR #61, opened 2016-02-29
feat(domains_tests): add domains tests

vdice:


tests/domains_test.go
```diff
@@ -1,21 +1,134 @@
...
-			// "deis domains:remove %s --app=%s", domain, app
-			// "deis domains:list --app=%s", app
 			// curl app at both root and custom domain, custom should fail
+			curlCmd = Cmd{CommandLineString: fmt.Sprintf(`curl -sL -w "%%{http_code}\\n" "%s" -o /dev/null`, testApp.URL)}
+			Eventually(cmdWithRetry(curlCmd, strconv.Itoa(200), cmdRetryTimeout)).Should(BeTrue())
+			curlCmd = Cmd{CommandLineString: fmt.Sprintf(`curl -sL -H "Host: %s" -w "%%{http_code}\\n" "%s" -o /dev/null`, domain, testApp.URL)}
+			Eventually(cmdWithRetry(curlCmd, strconv.Itoa(404), cmdRetryTimeout)).Should(BeTrue())
 		})
 
 		XIt("can add, list, and remove certs", func() {
```
vdice: Note: `certs` tests still need to be written.  Having trouble finding a resource on how this functionality is supposed to look; tempted to create a separate ticket to fill these out as I'll at least need assistance from a knowledgeable resource...

helgi: https://github.com/deis/workflow/blob/master/rootfs/api/tests/test_certificate.py gets you part way, doesn't tell you how the commands interact per se but it has majority of the required scaffolding you'd need to write

helgi: actually all the certificate ones in https://github.com/deis/workflow/tree/master/rootfs/api/tests

helgi: btw there is a separate ticket for certificates. I'd suggest moving tthat work there and delete this part of the code; You are going to need domains in certs but whatever, granularity and all that. Certs can stand on their own if you aren't going to be hitting an endpoint

vdice: #32 right?  Ok, I think that's a good idea.  I was just playing around with the tests, will put comments/questions on that ticket.

helgi: That's the one! Makes it easier to reason with since Certificates are considered a top level resources

tests/domains_test.go
```diff
@@ -1,21 +1,134 @@
...
+			cmd = createApp(testApp.Name)
+			Eventually(cmd).Should(SatisfyAll(
+				Say("Git remote deis added"),
+				Say("remote available at ")))
+		})
+
+		AfterEach(func() {
+			destroyApp(testApp)
+			// TODO: gitClean()
+			cmd, err := start("rm -rf .git")
```
bacongobbler: `git remote rm deis`?

vdice: Yes, that would also work.  I just pulled over the `rm -rf .git` block from another test and wrote a TODO more as a reminder to transfer that and the `git init` logic into re-usable utils (to be placed in `tests_suite_test` for now, since we don't have a dedicated `utils` resource as of yet.)  How does that sound?  Then any test could simply call `gitInit()` in setup and `gitClean()` in teardown

vdice: Have updated commit with idea mentioned above...

tests/domains_test.go
```diff
@@ -1,34 +1,128 @@
...
+
+			sess, err = start("deis domains:list --app=%s", testApp.Name)
+			Expect(err).NotTo(HaveOccurred())
+			Eventually(sess).Should(Say("=== %s Domains", testApp.Name))
+			Eventually(sess).Should(Say("%s", domain))
+			Eventually(sess).Should(Exit(0))
+
+			// curl app at both root and custom domain, both should return 200
+			curlCmd = Cmd{CommandLineString: fmt.Sprintf(`curl -sL -w "%%{http_code}\\n" "%s" -o /dev/null`, testApp.URL)}
+			Eventually(cmdWithRetry(curlCmd, strconv.Itoa(200), cmdRetryTimeout)).Should(BeTrue())
```
bacongobbler: if we're expecting status codes, should we use `http.StatusOK` instead of 200 to indicate that? https://golang.org/pkg/net/http/#pkg-constants

vdice: Good call; changed.

krancour:
The tests look ok to me, but I am wondering why we have a restriction on adding or removing domains before the app has been deployed.  That's not immediately obvious to me.  cc @helgi

tests/domains_test.go
```diff
@@ -1,34 +1,129 @@
...
+
+		It("can list domains", func() {
+			sess, err := start("deis domains:list --app=%s", testApp.Name)
+			Expect(err).NotTo(HaveOccurred())
+			Eventually(sess).Should(Say("=== %s Domains", testApp.Name))
+			Eventually(sess).ShouldNot(Say("%s", testApp.Name))
+			Eventually(sess).Should(Exit(0))
+		})
+
+		It("cannot add or remove domains", func() {
```
helgi: Is this for when you have yet to do a deploy?

tests/domains_test.go
```diff
@@ -1,34 +1,129 @@
...
+			Expect(err).NotTo(HaveOccurred())
+			Eventually(sess).Should(Say("Removing %s from %s...", domain, testApp.Name))
+			Eventually(sess).Should(Say("done"))
+			Eventually(sess).Should(Exit(0))
+
+			// attempt to remove non-existent domain
+			sess, err = start("deis domains:remove %s --app=%s", domain, testApp.Name)
+			Expect(err).NotTo(HaveOccurred())
+			Eventually(sess.Err).Should(Say("500 Internal Server Error")) // better error/explanation needed from cli
+			Eventually(sess).Should(Exit(1))
```
helgi: Ugh that's an annoying problem. A 500 isn't a good thing here but well... can you log a ticket because I think we should not bomb out with 500 on this scenario, a 400 maybe or 4xx of some sort

tests/domains_test.go
vdice: Yes

helgi: Okay cool - in upcoming PRs (we should merge this one first) I am going to remove the need for that

tests/domains_test.go
vdice: https://github.com/deis/workflow/issues/440

PR #67, opened 2016-03-03
feat(tags_tests.go): implement tags-related tests

### LINKED ISSUE #18: write the tags tests
they're all commented out: https://github.com/deis/workflow-e2e/blob/master/tests/tags_test.go

mboersma:
~~Still a couple of `// TODO`s and a 500 error to investigate.~~

~~I think deis-workflow's validation on labels (tags) needs to be improved to match https://github.com/kubernetes/kubernetes/blob/master/docs/user-guide/labels.md#syntax-and-character-set. Currently we don't allow `deis tags:set myprefix/mytag=myvalue` because we disallow a slash `/` in the tag name.~~ Fixed in #481.

Closes #18.

### DIFF
--- a/tests/tags_test.go
+++ b/tests/tags_test.go
@@ -1,16 +1,146 @@
 package tests
 
 import (
+	"os"
+	"os/exec"
+	"regexp"
+	"strings"
+	"sync"
+
 	. "github.com/onsi/ginkgo"
-	// . "github.com/onsi/gomega"
+	. "github.com/onsi/ginkgo/extensions/table"
+	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gbytes"
+	. "github.com/onsi/gomega/gexec"
 )
 
 var _ = Describe("Tags", func() {
+
 	Context("with a deployed app", func() {
+		var testApp App
+		once := &sync.Once{}
 
-		// TODO: does "deis tag" have a k8s implementation?
-		XIt("can set a tag", func() {
+		BeforeEach(func() {
+			// use the "kubectl" executable in the search $PATH
+			if _, err := exec.LookPath("kubectl"); err != nil {
+				Skip("kubectl not found in search $PATH")
+			}
 
+			// Set up the Tags test app only once and assume the suite will clean up.
+			once.Do(func() {
+				os.Chdir("example-go")
+				appName := getRandAppName()
+				createApp(appName)
+				testApp = deployApp(appName)
+			})
+		})
+
+		It("can set and unset tags", func() {
+			// can list tags
+			sess, err := start("deis tags:list")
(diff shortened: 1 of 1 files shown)


helgi:
I believe only 1x `/` is allowed

helgi:
https://github.com/deis/workflow/blob/master/rootfs/api/serializers.py#L179-L190 is where the validation logic happens

mboersma:
This is all working well locally after deis/workflow#481, but there still needs to be a way to query the k8s cluster to find legal values for `kubernetes.io/hostname` labels. The hard-coded value here works on kube-solo for me, but probably only there...

helgi:
You could do something like `k get nodes 172.17.8.100 --template={{.metadata.labels}}` - Returns a `map` but does require you to know about the node you want to work against

mboersma:
> You could do something like `k get nodes`

I tried that same approach in a previous version of this PR, but the trick is getting access to `kubectl` from within the test environment with a fake `$HOME` dir and other complications. That's what I'm trying to figure out currently.

mboersma:
This should work now on any k8s environment with node tags, but I could use help verifying that.

tests/tags_test.go
```diff
@@ -1,16 +1,146 @@
...
-		// TODO: does "deis tag" have a k8s implementation?
-		XIt("can set a tag", func() {
+			// find a valid tag to set
+			cmd := "kubectl get nodes -o jsonpath={.items[*].metadata..labels}"
+			// use original $HOME dir or kubectl can't find its config
+			sess, err = start("HOME=%s %s", homeHome, cmd)
+			Expect(err).NotTo(HaveOccurred())
+			Eventually(sess).Should(Exit(0))
+			// grep output like "map[kubernetes.io/hostname:192.168.64.2 node:worker1]"
+			re := regexp.MustCompile(`([\w\.]{0,253}/?[-_\.\w]{1,63}:[-_\.\w]{1,63})`)
```
bacongobbler: Will these tests fail if we haven't set any node tags? Just wondering if we need to update the documentation in https://github.com/deis/workflow-e2e#set-up-a-deis-cluster so testers know how to set up the cluster and make the tests pass.

mboersma: I think that all cluster nodes should have at least the `kubernetes.io/hostname` label set, which satisfies this test. 

I could move this code into the `BeforeEach` and have the whole spec get `Skip`ped if no labels actually exist, if you think that's better.

bacongobbler: nah the default hostname label is good enough. :)

vdice:
Looks good!
```
Ran 13 of 116 Specs in 114.786 seconds
SUCCESS! -- 13 Passed | 0 Failed | 0 Pending | 103 Skipped PASS
```

PR #72, opened 2016-03-03
feat(domains_test): workflow #448 fixes behavior; test needed updating

vdice:


tests/domains_test.go
```diff
@@ -109,7 +109,7 @@ var _ = Describe("Domains", func() {
 			// attempt to remove non-existent domain
 			sess, err = start("deis domains:remove %s --app=%s", domain, testApp.Name)
 			Expect(err).NotTo(HaveOccurred())
```
bacongobbler: Should an error be expected here instead? `Expect(err).To(HaveOccurred())`

bacongobbler: oh this is not what I was expecting. I see line 113 below. carry on!

PR #77, opened 2016-03-04
fix(jobs): prevent workflow-e2e-pr job from pushing canaries to registries

sgoings:


vdice:
Also, `_scripts/deploy.sh` could take in the type of docker-push as its first/only argument: `deploy.sh [mutable/immutable]-docker-push`, or maybe pass a flag... just ideas to keep DRY.

sgoings:
@vdice yeah... I get where you're coming from... however I see a future fairly soon in workflow-e2e where we don't need the deploy.sh script (which is only necessary to be a bash script because of the limitations of TravisCI). Since workflow-e2e is quickly moving to run all within Jenkins (for PRs and such) I didn't see a need to give deploy.sh any love. :smile:

This was a first pass at a quick fix (which is preventing deis-e2e tests from having consistent and understandable behavior) and then I was going to take a crack at:

* removing the travis job for workflow-e2e (because now we have a Travis job and a Jenkins job that do very similar things)
* porting the deploy.sh logic into the Makefile and then deleting deploy.sh (in this repo)

vdice:
roger; good plan!

PR #82, opened 2016-03-04
fix(version_test): version now includes 7-character git sha

### LINKED ISSUE #80: Recent change in versioning on the workflow client broke tests
```
• Failure [1.593 seconds]
Version
/go/src/github.com/deis/workflow-e2e/tests/version_test.go:16
  prints its version [It]
  /go/src/github.com/deis/workflow-e2e/tests/version_test.go:15

  Expected
      <string>: 2.0.0-dev-916fe94

  to equal
      <string>: 2.0.0-dev
```

vdice:
Fixes #80

### DIFF
--- a/tests/version_test.go
+++ b/tests/version_test.go
@@ -6,11 +6,12 @@ import (
 )
 
 var _ = Describe("Version", func() {
+	ourSemVerRegExp := `^v?(?:[0-9]+\.){2}[0-9]+`
+	gitSHARegExp := `[0-9a-f]{7}`
 
 	It("prints its version", func() {
 		output, err := execute("deis --version")
 		Expect(err).NotTo(HaveOccurred())
-		// TODO: read the expected version from ../client/deis-version?
-		Expect(output).To(Equal("2.0.0-dev\n"))
+		Expect(output).To(MatchRegexp(`%s(-dev)?(-%s)?\n`, ourSemVerRegExp, gitSHARegExp))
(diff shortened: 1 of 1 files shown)


tests/version_test.go
```diff
@@ -10,7 +10,6 @@ var _ = Describe("Version", func() {
 	It("prints its version", func() {
 		output, err := execute("deis --version")
 		Expect(err).NotTo(HaveOccurred())
-		// TODO: read the expected version from ../client/deis-version?
```
vdice: this todo didn't appear valid anymore

tests/version_test.go
```diff
@@ -10,7 +10,6 @@ var _ = Describe("Version", func() {
...
-		Expect(output).To(Equal("2.0.0-dev\n"))
+		Expect(output).To(MatchRegexp("2.0.0-dev-[0-9a-f]{7}\n"))
```
sgoings: Why not do a regex that reads something like:

3 digit, dot separated version
maybe a -dev suffix with a -(git sha with max 7 digits)

sgoings: Also:

"The rest is left as an exercise to the reader"

bacongobbler: Also the dots indicate any character. Here's a regex based on @sgoings suggestion:

```
^\d\.\d\.\d(-dev-\w{7})?\n$
```

PR #99, opened 2016-03-10
fix(Dockerfile): pass same options to test binary

### LINKED ISSUE #97: Pass `make test-integration` params to test binary
When the deis-tests chart is run, it executes `/bin/tests.test` inside a container. This ignores the Ginkgo command-line flags used in `make test-integration`, so many "slow test" warnings are printed, for example.

Either `tests.test` should somehow accept the same command-line options, allow them to be baked in to the binary, or the Docker image for workflow-e2e should instead just call `make test-integration` internally.

mboersma:
~~Closes #97.~~

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,8 @@ DEV_IMG := quay.io/deis/go-dev:0.9.0
 RUN_CMD := docker run --rm -e DEIS_ROUTER_SERVICE_HOST=${DEIS_ROUTER_SERVICE_HOST} -e DEIS_ROUTER_SERVICE_PORT=${DEIS_ROUTER_SERVICE_PORT} -v ${CURDIR}:${SRC_PATH} -w ${SRC_PATH} ${DEV_IMG}
 DEV_CMD := docker run --rm -e GO15VENDOREXPERIMENT=1 -v ${CURDIR}:${SRC_PATH} -w ${SRC_PATH} ${DEV_IMG}
 
+TEST_OPTS := -v -timeout=60m -ginkgo.v -ginkgo.slowSpecThreshold=120
+
 MUTABLE_VERSION ?= canary
 VERSION ?= git-$(shell git rev-parse --short HEAD)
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -4,4 +4,4 @@ COPY tests/tests.test .
 RUN mv tests.test /bin
 RUN apt-get update -y && apt-get install -y curl openssh-client git
 RUN curl -sSL http://deis.io/deis-cli/install-v2-alpha.sh | bash && mv ./deis /bin/deis
-CMD /bin/tests.test
+CMD ["/bin/tests.test", "-v", "-timeout=60m", "-ginkgo.v", "-ginkgo.slowSpecThreshold=120"]
```
vdice: Would it be possible to move these into the Makefile as args to the `docker run` command issued by the `docker-test-integration` target? Something like:
```
	docker run -e DEIS_ROUTER_SERVICE_HOST=${DEIS_ROUTER_SERVICE_HOST} -e DEIS_ROUTER_SERVICE_PORT=${DEIS_ROUTER_SERVICE_PORT} ${IMAGE} ${TEST_OPTS}
```
If that would work, then we could have one canonical version (in the Makefile) of these options that both targets could consume.

mboersma: > we could have one canonical version (in the Makefile) of these options that both targets could consume.

Yes, much better idea. Let me give that a go.

mboersma: I guess by baking the args into the default Dockerfile `CMD`, I was hoping this would Just Work&trade; for the deis-tests chart. But it appears I will have to make similar changes [over there](https://github.com/deis/charts/blob/master/deis-tests/manifests/deis-tests-pod.yaml#L13-L14).

mboersma:
Updated based on a suggestion by @vdice. I realized I will have to make a separate PR to close #97 however, since the command and options are part of the deis-tests chart.

vdice:
I see.  I'd definitely be in favor of baking it into the Dockerfile/image if it helped multiple consumers (`deis-tests` chart being the prime example) and thereby sacrifice a small bit of DRYness here in this repo... whichever is the best way to align how the e2e tests are run.  All said, LGTM3

PR #103, opened 2016-03-15
fix(certs): this brings the necessary certs files into proper place for tests container

vdice:


tests/tests_suite_test.go
```diff
@@ -44,8 +44,12 @@ var (
 )
 
 func getDir() string {
-	_, filename, _, _ := runtime.Caller(1)
-	return path.Dir(filename)
+	if os.Getenv("HOME") != "/root" {
```
vdice: Definitely open to other suggestions on better ways to determine if we're in a Docker container vs. on a dev's box...

Makefile
```diff
@@ -8,7 +8,7 @@ DEV_IMG := quay.io/deis/go-dev:0.9.0
 RUN_CMD := docker run --rm -e DEIS_ROUTER_SERVICE_HOST=${DEIS_ROUTER_SERVICE_HOST} -e DEIS_ROUTER_SERVICE_PORT=${DEIS_ROUTER_SERVICE_PORT} -v ${CURDIR}:${SRC_PATH} -w ${SRC_PATH} ${DEV_IMG}
 DEV_CMD := docker run --rm -e GO15VENDOREXPERIMENT=1 -v ${CURDIR}:${SRC_PATH} -w ${SRC_PATH} ${DEV_IMG}
 
-TEST_OPTS := -v -timeout=60m -ginkgo.v -ginkgo.slowSpecThreshold=120
+TEST_OPTS := -test.v -test.timeout=60m -ginkgo.v -ginkgo.slowSpecThreshold=120
```
mboersma: This change from deis/charts#151 isn't strictly necessary here since these flags are given directly to `go test`. But I think this will act the same as the shorter versions (need to confirm!), and I would prefer this change that makes things more consistent.

tests/tests_suite_test.go
mboersma: Testing for the presence of a `.dockerinit` (or `.dockerenv`) file is one way to know if you're inside a container.
```go
if _, err := os.Stat("/.dockerinit"); err == nil {
```

tests/tests_suite_test.go
```diff
@@ -44,6 +44,10 @@ var (
 )
 
 func getDir() string {
+	// if inside Docker container, don't return nothin'
```
arschles: Is it worth specifying we're inside a container explicitly with an env var? For example, the minio project uses `DOCKERIMAGE=1` to indicate that it's running inside a container

arschles: ^^ not blocking for this PR

arschles:
@vdice code LGTM. I raised one question at https://github.com/deis/workflow-e2e/pull/103#discussion_r56207392, but not blocking for this PR

tests/tests_suite_test.go
vdice: @mboersma talked about this option at one point, not really having a strong preference. Is it a better practice?

arschles: My preference is to follow [12-factor](http://12factor.net/) practices and specify it in an env var.

mboersma: @arschles makes a good point about `DOCKERIMAGE`. I guess testing for `.dockerinit` is really relying on an implementation detail of Docker that may change someday. 

Sorry for the short-sighted advice. :crying_cat_face:

PR #115, opened 2016-03-16
feat(perms_test): we do not need to deploy the app for current perms tests

vdice:


arschles:
I accidentally removed @sgoings's `LGTM1` while I added my `LGTM2`. Re-added it. Concurrency is hard....

PR #118, opened 2016-03-18
fix(releases_test): rollback to actual build+config release in test

vdice:
Previously we were rolling back to v1 which isn't a legit build+config/12-factor
release. cc @bacongobbler @mboersma 

Closes https://github.com/deis/controller/issues/538

### DIFF
--- a/tests/builds_test.go
+++ b/tests/builds_test.go
@@ -15,12 +15,23 @@ import (
 // createBuild invokes deis builds:create <image> -a <app>
 // with provided <image> on provided <app>
 // and validates that no errors have occurred and build was successful
-func createBuild(image string, testApp App) {
-	cmd, err := start("deis builds:create %s -a %s", image, testApp.Name)
-	Expect(err).NotTo(HaveOccurred())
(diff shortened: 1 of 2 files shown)


tests/builds_test.go
```diff
@@ -15,12 +15,23 @@ import (
...
 // with provided <image> on provided <app>
 // and validates that no errors have occurred and build was successful
-func createBuild(image string, testApp App) {
-	cmd, err := start("deis builds:create %s -a %s", image, testApp.Name)
-	Expect(err).NotTo(HaveOccurred())
-	Eventually(cmd, defaultMaxTimeout).Should(Exit(0))
-	Eventually(cmd).Should(Say("Creating build..."))
-	Eventually(cmd).Should(Say("done"))
+func createBuild(image string, app App, options ...string) {
+	pullORCreateBuild(image, app, "builds:create", strings.Join(options, " "))
```
mboersma: Shouldn't this identifier by `pullOrCreateBuild`?

vdice: Indeed; updated

PR #121, opened 2016-03-22
Add long running tests

kmala:
Add long running tests which does continuous deployments in parallel.


Makefile
```diff
@@ -9,6 +9,7 @@ RUN_CMD := docker run --rm -e DEIS_ROUTER_SERVICE_HOST=${DEIS_ROUTER_SERVICE_HOS
 DEV_CMD := docker run --rm -e GO15VENDOREXPERIMENT=1 -v ${CURDIR}:${SRC_PATH} -w ${SRC_PATH} ${DEV_IMG}
 
 TEST_OPTS := -test.v -test.timeout=60m -ginkgo.v -ginkgo.slowSpecThreshold=120
+LONGRUN_TEST_OPTS := -test.v -ginkgo.v -ginkgo.slowSpecThreshold=120 -test.parallel=4
```
vdice: This doesn't appear to be used anywhere

shims/system.go
```diff
@@ -10,10 +10,14 @@ import (
 type Shim struct {
 	OutFile  *os.File
 	ShimFile *os.File
+	Dir      string
```
vdice: :+1: these changes definitely improve the robustness of this utility

longruntests/longruntests_suite_test.go
```diff
@@ -0,0 +1,404 @@
+package longruntests_test
```
vdice: Is this file a wholesale copy of https://github.com/deis/workflow-e2e/blob/master/tests/tests_suite_test.go ?  if so, can we just use that?  

If no, can we just add any `longrun`-specific functions into the aforementioned file?  If the majority of the code is identical, definitely want to see it in one place only...

utils/utils.go
```diff
@@ -0,0 +1,478 @@
...
+		Login(url, username, password)
+	} else {
+		Eventually(sess).Should(Exit(0))
+		Eventually(sess).Should(SatisfyAll(
+			Say("Registered %s", username),
+			Say("Logged in as %s", username)))
+	}
+}
+
+// execute executes the command generated by fmt.Sprintf(cmdLine, args...) and returns its output as a cmdOut structure.
```
vdice: nit: now that some of these methods have now been capitalized (officially public), the godoc description (comments) should be updated as well.

longruntests/longruntests_suite_test.go
```diff
@@ -0,0 +1,152 @@
...
+	. "github.com/deis/workflow-e2e/utils"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gbytes"
+	. "github.com/onsi/gomega/gexec"
+
+	"github.com/onsi/ginkgo/reporters"
+)
+
+func getDir() string {
```
vdice: I'd still like to see if it would be possible to have all the code duplicated here (`getDir()`, `init()`, `getRandAppName()`, `TestTests`, `var` listing, etc.) exist in only one file where possible (probably `tests_suite_test.go` or the new `utils`).

longruntests/longruntests_suite_test.go
arschles: Also, can we not use `_test` in package names? Everything in this repo is tests, so I think this suffix is redundant

longruntests/longruntests_suite_test.go
```diff
@@ -0,0 +1,152 @@
...
+	"io/ioutil"
+	"math/rand"
+	"os"
+	"os/exec"
+	"path"
+	"runtime"
+	"testing"
+	"time"
+
+	. "github.com/deis/workflow-e2e/utils"
```
arschles: I'd rather not anonymous import this package. The below ones are fine, as we've settled on that convention in this repo. Let's not expand it further, though.

tests/apps_test.go
```diff
@@ -1,22 +1,16 @@
 package tests
 
 import (
-	"io/ioutil"
 	"os"
-	"runtime"
-	"strings"
 	"time"
 
+	. "github.com/deis/workflow-e2e/utils"
```
arschles: same here regarding the anonymous import

tests/apps_test.go
```diff
@@ -181,67 +175,24 @@ var _ = Describe("Apps", func() {
...
-	// invoke functionality under test
-	sess, err := startCmd(Cmd{Env: env, CommandLineString: "deis open"})
-	Expect(err).To(BeNil())
-	Eventually(sess).Should(Exit(0))
-
-	// check shim output
-	output, err := ioutil.ReadFile(myShim.OutFile.Name())
-	Expect(err).NotTo(HaveOccurred())
-	Expect(strings.TrimSpace(string(output))).To(Equal(testApp.URL))
-}
```
arschles: was this code moved?

tests/config_test.go
```diff
@@ -5,6 +5,7 @@ import (
 	"os"
 	"sync"
 
+	. "github.com/deis/workflow-e2e/utils"
```
arschles: same here regarding the anonymous import

tests/domains_test.go
```diff
@@ -7,6 +7,7 @@ import (
 	"os"
 	"strconv"
 
+	. "github.com/deis/workflow-e2e/utils"
```
arschles: same here regarding the anonymous import

tests/keys_test.go
```diff
@@ -6,6 +6,7 @@ import (
 	"os"
 	"time"
 
+	. "github.com/deis/workflow-e2e/utils"
```
arschles: same here regarding the anonymous import

arschles: and ones below, I won't comment on all of them

utils/utils.go
```diff
@@ -0,0 +1,478 @@
...
+}
+
+func Logout() {
+	sess, err := Run("deis auth:logout")
+	Expect(err).To(BeNil())
+	Eventually(sess).Should(Exit(0))
+	Eventually(sess).Should(Say("Logged out\n"))
+}
+
+func Run(cmdLine string, args ...interface{}) (*Session, error) {
```
arschles: I get that this is now exported, but why rename this to `Run` (instead of `Start`)?

tests/apps_test.go
kmala: Yes.to the utils/utils.go

utils/utils.go
kmala: gexec already has a function with name Start

arschles:
I commented with a bunch of mostly-nitpick comments, but generally speaking, this looks good. General question: are these long-running tests going to share a codebase with the "normal" e2e tests? I ask because I want to make sure everything herein can be run as a short-running _and_ long running test.

longruntests/longruntests_suite_test.go
kmala: I had used it since they are already used in the repo. I am happy to change them but just wanted to know the reason for not using the anonymous  import.

longruntests/longrun_test.go
```diff
@@ -0,0 +1,200 @@
...
+		time.Sleep(5 * time.Second)
+	}
+}
+
+var _ = Describe("Longrun", func() {
+
+	It("Run example Go app", func() {
+		var testApp App
+		os.Chdir("example-go")
+		for {
```
vdice: Duplication could be vastly reduced here.  One approach would to be set these up in a table-driven style; something like [this](https://gist.github.com/vdice/213ebdfda93289e3e9518557cf3a3b09)

kmala: The table-driven style looks good...I'll try that and update the PR

longruntests/longruntests_suite_test.go
```diff
@@ -0,0 +1,138 @@
...
+
+var _ = BeforeEach(func() {
+	var err error
+	var output string
+
+	testRoot, err = ioutil.TempDir("", "deis-workflow-test")
+	Expect(err).NotTo(HaveOccurred())
+
+	os.Chdir(testRoot)
+	output, err = Execute(`git clone https://github.com/deis/example-go.git`)
```
vdice: nit: might be better to create a data structure for all of the example app names (`["example-go", "example-php",...]`) and then iterate over all of these with the same `output, err = ..., Expect(...` logic.  That way its nice and easy to add another example app at a later date.  

Also, if this route is taken, other files can use said canonical exampleApps data structure, such as `longrun_test.go` (If the table-driven approach mentioned in that file is pursued, you could iterate over `exampleApps` for the creation of each `Entry(appName)`)

vdice:
@kmala where is this at?  Assuming it won't make it to rc1?  If not, please move to appropriate milestone.

jchauncey:
Can we jsut close this PR for now?

PR #131, opened 2016-03-25
fix(ps): Expect 503 when app scaled to 0

### LINKED ISSUE #127: When app is scaled to zero, should assert response is 503 (not 502)

krancour:
Fixes #127

Complements deis/router#153.  This PR adjusts the expected response code when an application is scaled to 0-- i.e. when there are no pods backing an upstream service.

Apart from the altered expectation, this also removes the specs in question from the pending state.  As such, this should be merged _after_ deis/router#153 to try to avoid a failure in CI.


vdice:
@krancour although some ps tests are still marked as `Pending`, the expectations have been updated to assert `503` now: https://github.com/deis/workflow-e2e/blob/master/tests/ps_test.go

Therefore, should this PR be closed?

jchauncey:
@krancour is this still a valid PR?

krancour:
@jchauncey nope.  This change was included in the refactor.  Closing.

PR #135, opened 2016-03-28
feature: Run the e2e test suite in parallel

jchauncey:
This PR gives us the ability to run the e2e test suite in parallel. This dramatically speeds up the test run making it easier on the end user who is trying to get a crucial change. For example here is a sequential test run performed against my 4 node m4.xlarge cluster -

```
Ran 100 of 113 Specs in 2003.951 seconds
SUCCESS! -- 100 Passed | 0 Failed | 13 Pending | 0 Skipped PASS
```

That test run took almost 33 minutes to complete.

But when running the tests in parallel we can see results like the following:

```
Ran 100 of 113 Specs in 367.949 seconds
SUCCESS! -- 100 Passed | 0 Failed | 13 Pending | 0 Skipped
```
That is just over 6 minutes.

To accomplish this several changes were made to the test suite:

1. Use http to register and login all users. This allows us to store the login token in a `user.json` file within the `$TEST_HOME` directory. 

1. Use `DEIS_PROFILE` to ensure we use the correct user for each action. When running tests in parallel we cannot be certain that the right fork/exec of the deis cli command is using the correct logged in user. 

1. Removed (commented out) the reset all user tokens test. So this test was wreaking havoc on my test runs because it basically blanked out all login tokens I had generated up until that point. This meant other tests that were inflight would suddenly start getting 401's. Until we can figure out a better way to test this I suggest we leave it commented out.

1.  Made most tests use the `Eventually` matcher. This is more of a consistency thing than for the parallel tests.

1. Removed lots of clean up code. Because we reset the cluster after each run and each test generates its own data we no longer need to cleanup after ourselves. The one notable exception to this is the domains/certs tests as I wasnt sure if we needed to clean up between tests for them to behave properly. However, I am pretty sure that I can remove that as well @helgi, @mboersma?

1.  `defaultMaxTimeout` is now at 600 seconds and `DefaultEventuallyTimeout` is 60 seconds. For most of last week I was dealing with timeouts when running these tests in parallel. I eventually found out that my cluster was mostly to blame. That being said I found having a longer timeout eased some of the failures I was seeing.

1. Use `fmt.Fprintf(GinkgoWriter...)`. Because we are running tests in parallel we can no longer use the standard `fmt.Printf` or `fmt.Println` calls because the output is jumbled. The `GinkgoWriter` is setup to handle parallel test runs.

1. Use go-dev as base image for docker and quit building the tests binary. The reasoning behind this is that the test binary doesnt allow us to run the tests in parallel. That means we have to add the tests to the image and use the regular `ginkgo` command to execute them.

1. Made the `defaultMaxTimeout` and `DefaultEventuallyTimeout` configurable by environment variables.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,18 @@
-FROM ubuntu-debootstrap:14.04
+FROM quay.io/deis/go-dev:0.9.0
 
-ENV DOCKERIMAGE=1
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends ca-certificates curl wget openssh-client git g++ gcc libc6-dev make bzr git mercurial openssh-client subversion procps && \
+    rm -rf /var/lib/apt/lists/*
 
-RUN apt-get update -y && apt-get install -y curl openssh-client git
-RUN curl -sSL http://deis.io/deis-cli/install-v2-alpha.sh | bash && mv ./deis /bin/deis
-COPY tests/tests.test .
-RUN mv tests.test /bin
-RUN mkdir /files
-COPY tests/files /files
-CMD /bin/tests.test
+#Install deis cli
+RUN wget https://dl.bintray.com/deis/deisci/deis-6c176d2-linux-amd64 && \
+    mv ./deis-6c176d2-linux-amd64 /bin/deis && \
+    chmod +x /bin/deis
+
+# Install the kubectl cli
+RUN curl -O https://storage.googleapis.com/kubernetes-release/release/v1.1.8/bin/linux/amd64/kubectl && \
+    chmod +x kubectl && \
+    mv kubectl /usr/local/bin/kubectl
+
+COPY . /go/src/github.com/deis/workflow-e2e
+WORKDIR /go/src/github.com/deis/workflow-e2e
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ bootstrap:
 
 .PHONY: test-integration
 test-integration:
-	go test ./tests/... ${TEST_OPTS}
+	ginkgo tests/ -p -slowSpecThreshold=120.00 -succinct -progress
 
 # Precompile the test suite into a binary "_tests.test"
 .PHONY: build
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,8 @@ DEV_IMG := quay.io/deis/go-dev:0.9.0
 RUN_CMD := docker run --rm -e DEIS_ROUTER_SERVICE_HOST=${DEIS_ROUTER_SERVICE_HOST} -e DEIS_ROUTER_SERVICE_PORT=${DEIS_ROUTER_SERVICE_PORT} -v ${CURDIR}:${SRC_PATH} -w ${SRC_PATH} ${DEV_IMG}
 DEV_CMD := docker run --rm -e GO15VENDOREXPERIMENT=1 -v ${CURDIR}:${SRC_PATH} -w ${SRC_PATH} ${DEV_IMG}
 
-TEST_OPTS := -test.v -test.timeout=60m -ginkgo.v -ginkgo.slowSpecThreshold=120
+TEST_OPTS := -slowSpecThreshold=120.00 -noisyPendings=false
+PARALLEL_TEST_OPTS := ${TEST_OPTS} -p
 
 MUTABLE_VERSION ?= canary
 VERSION ?= git-$(shell git rev-parse --short HEAD)
@@ -24,15 +25,14 @@ bootstrap:
 
 .PHONY: test-integration
 test-integration:
-	ginkgo tests/ -p -slowSpecThreshold=120.00 -succinct -progress
+	DEFAULT_EVENTUALLY_TIMEOUT="30s" ginkgo $TEST_OPTS tests/
 
-# Precompile the test suite into a binary "_tests.test"
-.PHONY: build
-build:
-	${DEV_CMD} ginkgo build -race -r
+.PHONY: test-integration
+test-integration-parallel:
+	ginkgo $PARALLEL_TEST_OPTS tests/
 
 .PHONY: docker-build
-docker-build: build
+docker-build:
 	docker build -t ${IMAGE} ${CURDIR}
 	docker tag -f ${IMAGE} ${MUTABLE_IMAGE}
 
@@ -51,5 +51,13 @@ docker-mutable-push:
 # run tests inside of a container
 docker-test-integration:
 	docker run -e DEIS_ROUTER_SERVICE_HOST=${DEIS_ROUTER_SERVICE_HOST} \
-	  -e DEIS_ROUTER_SERVICE_PORT=${DEIS_ROUTER_SERVICE_PORT} ${IMAGE} \
-	  /bin/tests.test ${TEST_OPTS}
+	  				 -e DEIS_ROUTER_SERVICE_PORT=${DEIS_ROUTER_SERVICE_PORT} \
+						 -e TEST_OPTS=${TEST_OPTS}
+						 -e DEFAULT_EVENTUALLY_TIMEOUT="30s" ${IMAGE}
+
+.PHONY: docker-test-integration-parallel
+# run tests inside of a container
+docker-test-integration-parallel:
+	docker run -e DEIS_ROUTER_SERVICE_HOST=${DEIS_ROUTER_SERVICE_HOST} \
+	  				 -e DEIS_ROUTER_SERVICE_PORT=${DEIS_ROUTER_SERVICE_PORT} \
+						 -e TEST_OPTS=${PARALLEL_TEST_OPTS} ${IMAGE}
--- a/tests/admin_test.go
+++ b/tests/admin_test.go
@@ -1,0 +1,96 @@
+package tests
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gbytes"
+	. "github.com/onsi/gomega/gexec"
+)
+
+var _ = Describe("Admin", func() {
+	Context("when logged in as an admin user", func() {
+		var testData TestData
+		var testApp App
+
+		BeforeEach(func() {
+			testData = initTestData()
+			testApp.Name = getRandAppName()
+			gitInit()
+			createApp(testData.Profile, testApp.Name)
+		})
+
+		It("can create, list, and delete admin permissions", func() {
+			sess, err := start("deis perms:create %s --admin", adminTestData.Profile, testData.Username)
+			Eventually(sess, defaultMaxTimeout).Should(Say("Adding %s to system administrators... done\n", testData.Username))
+			Expect(err).NotTo(HaveOccurred())
+			Eventually(sess).Should(Exit(0))
+			Expect(err).NotTo(HaveOccurred())
+			sess, err = start("deis perms:list --admin", adminTestData.Profile)
+			Eventually(sess).Should(Say("=== Administrators"))
+			Eventually(sess).Should(Say(adminTestData.Username))
+			Eventually(sess).Should(Say(testData.Username))
+			Expect(err).NotTo(HaveOccurred())
+			Eventually(sess).Should(Exit(0))
+			Expect(err).NotTo(HaveOccurred())
+			sess, err = start("deis perms:delete %s --admin", adminTestData.Profile, testData.Username)
+			Eventually(sess, defaultMaxTimeout).Should(Say("Removing %s from system administrators... done", testData.Username))
+			Expect(err).NotTo(HaveOccurred())
+			Eventually(sess).Should(Exit(0))
+			Expect(err).NotTo(HaveOccurred())
+			sess, err = start("deis perms:list --admin", adminTestData.Profile)
+			Eventually(sess).Should(Say("=== Administrators"))
+			Expect(sess).ShouldNot(Say(testData.Username))
+			Expect(err).NotTo(HaveOccurred())
+			Eventually(sess).Should(Exit(0))
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		It("can create, list, and delete app permissions", func() {
+			sess, err := start("deis perms:create %s --app=%s", adminTestData.Profile, testData.Username, testApp.Name)
+			Eventually(sess, defaultMaxTimeout).Should(Say("Adding %s to %s collaborators... done\n", testData.Username, testApp.Name))
+			Expect(err).NotTo(HaveOccurred())
+			Eventually(sess).Should(Exit(0))
+			Expect(err).NotTo(HaveOccurred())
+			sess, err = start("deis perms:list --app=%s", adminTestData.Profile, testApp.Name)
+			Eventually(sess).Should(Say("=== %s's Users", testApp.Name))
+			Eventually(sess).Should(Say("%s", testData.Username))
+			Expect(err).NotTo(HaveOccurred())
+			Eventually(sess).Should(Exit(0))
+			Expect(err).NotTo(HaveOccurred())
+			sess, err = start("deis perms:delete %s --app=%s", adminTestData.Profile, testData.Username, testApp.Name)
+			Eventually(sess, defaultMaxTimeout).Should(Say("Removing %s from %s collaborators... done", testData.Username, testApp.Name))
+			Eventually(sess).Should(Exit(0))
+			Expect(err).NotTo(HaveOccurred())
+			sess, err = start("deis perms:list --app=%s", adminTestData.Profile, testApp.Name)
+			Eventually(sess).Should(Say("=== %s's Users", testApp.Name))
+			Eventually(sess).ShouldNot(Say("%s", testData.Username))
+			Expect(err).NotTo(HaveOccurred())
+			Eventually(sess).Should(Exit(0))
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		It("regenerates the token for a specified user", func() {
+			sess, err := start("deis auth:regenerate -u %s", adminTestData.Profile, testData.Username)
+			Eventually(sess).Should(Say("Token Regenerated"))
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		// This is marked pending because it resets all user auth tokens. Because we run the tests in parallel
+		// this can wreak havoc on tests that may be in flight. We will need to reevaluate how we want to test this functionality.
+		XIt("regenerates the token for all users", func() {
+			sess, err := start("deis auth:regenerate --all", adminTestData.Profile)
+			Eventually(sess).Should(Say("Token Regenerated"))
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		It("can list all users", func() {
+			sess, err := start("deis users:list", adminTestData.Profile)
+			Eventually(sess).Should(Say("=== Users"))
+			output := string(sess.Out.Contents())
+			Expect(output).To(ContainSubstring(adminTestData.Username))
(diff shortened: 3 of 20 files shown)


tests/certs_test.go
```diff
@@ -126,57 +128,52 @@ var _ = Describe("Certs", func() {
...
 		BeforeEach(func() {
+			testData = initTestData()
 			// Set up the test app only once and assume the suite will clean up.
-			once.Do(func() {
-				os.Chdir(exampleRepo)
-				appName := getRandAppName()
-				createApp(appName)
-				testApp = deployApp(appName)
-			})
+			os.Chdir(exampleRepo)
```
vdice: We can remove the comment above now that we are no longer only deploying once for the subsequent assertions

tests/tests_suite_test.go
```diff
@@ -249,7 +167,12 @@ func execute(cmdLine string, args ...interface{}) (string, error) {
 	return output, err
 }
 
-func start(cmdLine string, args ...interface{}) (*Session, error) {
+func start(cmdLine string, profile string, args ...interface{}) (*Session, error) {
+	if profile != "" {
+		envVars := append(os.Environ(), fmt.Sprintf("DEIS_PROFILE=%s", profile))
+		ourCommand := Cmd{Env: envVars, CommandLineString: fmt.Sprintf(cmdLine, args...)}
+		return startCmd(ourCommand)
+	}
```
vdice: nit: could simplify/reduce duplication a bit here w/ something like:
```
envVars := os.Environ()
if profile != "" {
  envVars = append(envVars,  fmt.Sprintf("DEIS_PROFILE=%s", profile))
}
ourCommand := Cmd{Env: envVars, CommandLineString: fmt.Sprintf(cmdLine, args...)}
return startCmd(ourCommand)

File: tests/tests_suite_test.go [hunk 3]
```diff
@@ -381,20 +304,101 @@ func cmdWithRetry(cmd Cmd, expectedResult string, timeout int) bool {
...
 // gitClean destroys the .git directory and verifies the command is successful
 func gitClean() {
-	cmd, err := start("rm -rf .git")
+	cmd, err := start("rm -rf .git", "")
 	Expect(err).NotTo(HaveOccurred())
 	Eventually(cmd).Should(Exit(0))
 }
+
+func registerHTTP(url string, username string, password string, email string) string {
+	registerURL := fmt.Sprintf("%s/v2/auth/register/", url)
```
vdice: If we plan on expanding the use of making API calls as done here and in `loginHTTP` below, it would be handy to have both of these (and future methods) invoke a helper function, something like:
```
func makeHTTPReq(url, method string, jsonData []byte) string {
  req, err := http.NewRequest(method, url, bytes.NewBuffer(jsonData))
  req.Header.Set("Content-Type", "application/json")

  client := &http.Client{}
  resp, err := client.Do(req)
  if err != nil {
    panic(err)
  }
  defer resp.Body.Close()

  respBody, _ := ioutil.ReadAll(resp.Body)
  return string(respBody)
}

File: tests/tests_suite_test.go [hunk 4]
```diff
@@ -57,6 +36,32 @@ type App struct {
...
+	Email         string
+	KeyName       string
+	KeyPath       string
+	Profile       string
+	ProfilePath   string
+	ControllerURL string
+}
+
+var adminTestData TestData
+var testRoot, testHome, keyPath, gitSSH string
```
vdice: I can't seem to find where `testHome` is actually set -- could you point me to the spot?

tests/tests_suite_test.go
```diff
@@ -381,20 +304,101 @@ func cmdWithRetry(cmd Cmd, expectedResult string, timeout int) bool {
...
+
+	client := &http.Client{}
+	resp, err := client.Do(req)
+	if err != nil {
+		panic(err)
+	}
+	defer resp.Body.Close()
+
+	loginBody, _ := ioutil.ReadAll(resp.Body)
+	token := strings.TrimSuffix(strings.Split(string(loginBody), ":")[1], "}")
```
vdice: A comment or two explaining this block of logic would be helpful.  (Looks like we are creating the deis client profile (json file), putting it in the proper place and making it executable?)  It might even be good to separate this logic out from `loginHTTP` (which would only be responsible for making the http req/getting resp) into another method, say, `createClientProfile`, which will do the steps below.

tests/tests_suite_test.go
```diff
@@ -381,20 +304,101 @@ func cmdWithRetry(cmd Cmd, expectedResult string, timeout int) bool {
...
+	email := fmt.Sprintf("test-%d@deis.io", randSuffix)
+	keyName := fmt.Sprintf("deiskey-%v", randSuffix)
+
+	registerHTTP(getController(), username, password, email)
+	profilePath := loginHTTP(getController(), username, password)
+
+	keyPath := createKey(username, keyName)
+	// Write out a git+ssh wrapper file to avoid known_hosts warnings
+	gitSSH = path.Join(testHome, username, ".ssh", "git-ssh")
+	sshFlags := "-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null"
```
vdice: Curious as to why these particular `sshFlags` were added?

tests/tests_suite_test.go
```diff
@@ -381,20 +304,101 @@ func cmdWithRetry(cmd Cmd, expectedResult string, timeout int) bool {
...
+	if debug {
+		sshFlags = sshFlags + " -v"
+	}
+	ioutil.WriteFile(gitSSH, []byte(fmt.Sprintf(
```
vdice: A comment on what this is doing would be helpful (and/or why the change is needed from how git SSH was previously set up.)

tests/tests_suite_test.go
jchauncey: https://github.com/deis/workflow-e2e/pull/135/files#diff-0c6502f3bbf7ad9e0b1001fc3c66f5efR187

tests/tests_suite_test.go
jchauncey: well you can basically only get the token via the login request so yeah you could make another method that did that piece but it would only be called from this one function.

tests/tests_suite_test.go
jchauncey: I copy and pasted this from where we initing the git-ssh file in another function. Not sure why they are used.

jchauncey: here it is on master - https://github.com/deis/workflow-e2e/blob/master/tests/tests_suite_test.go#L117

tests/tests_suite_test.go
jchauncey: see comment above

tests/tests_suite_test.go
jchauncey: k pushed this change

jchauncey:
This PR needs several more refactors to make the code more readable and idiomatic but Id like to see about getting it merged and refactor over the next several weeks so we can get the e2e runs under 10 minutes.

Makefile
```diff
@@ -26,6 +26,10 @@ bootstrap:
 test-integration:
 	go test ./tests/... ${TEST_OPTS}
 
+.PHONY: test-integration
+test-integration-parallel:
+	ginkgo -slowSpecThreshold=120.00 --nodes=30 tests/
```
mboersma: `ginkgo --nodes=30` looks equivalent to `go test -ginkgo.parallel.total=30`. If so, this could use `${TEST_OPTS}` as other Makefile targets do.

Also, why does this target test `tests/` but the one above it tests `./tests/...`? In this case they amount to the same thing, but they should be consistent.

tests/tests_suite_test.go
vdice: sorry, that supplied link doesn't seem to pinpoint the line ^^

tests/tests_suite_test.go
```diff
@@ -87,7 +92,7 @@ func TestTests(t *testing.T) {
 }
 
 var _ = BeforeSuite(func() {
-	SetDefaultEventuallyTimeout(10 * time.Second)
+	SetDefaultEventuallyTimeout(60 * time.Second)
```
vdice: I don't think I'm in favor of increasing this default timeout.  While the effects are less pronounced when running in parallel, it really slows down the sequential run.  Since we're relying on the sequential run for the time being, I vote we keep the previous 10 sec default and revisit at a later date.

mboersma:
After running tests using this PR, I'm left with 40 or so test-* namespaces and their pods. Now I have to write a little script just to clean up after the tests, which pretty much guarantees I'll never run the tests again if they do that. :-/

jchauncey:
for now just do `kg ns | sed 1d | awk '{print $1}' | grep test | xargs kubectl delete ns`

Dockerfile
```diff
@@ -1,11 +1,37 @@
...
-COPY tests/files /files
-CMD /bin/tests.test
+# RUN curl -sSL http://deis.io/deis-cli/install-v2-alpha.sh | bash && mv ./deis /bin/deis
+# COPY tests/
+# RUN mv tests.test /bin
+# RUN mkdir /files
+# COPY tests/files /files
+# CMD /bin/tests.test
+
+ENV GO15VENDOREXPERIMENT=1
```
vdice: If our base image is `go-dev` above, we shouldn't need lines 13-26 or lines 30-34 (https://github.com/deis/docker-go-dev/blob/master/rootfs/Dockerfile as reference)

jchauncey: good call. i think i copied them from that thinking i wouldnt use that image as a base but decided to anyways.

Makefile
```diff
@@ -51,5 +56,13 @@ docker-mutable-push:
 # run tests inside of a container
 docker-test-integration:
 	docker run -e DEIS_ROUTER_SERVICE_HOST=${DEIS_ROUTER_SERVICE_HOST} \
-	  -e DEIS_ROUTER_SERVICE_PORT=${DEIS_ROUTER_SERVICE_PORT} ${IMAGE} \
-	  /bin/tests.test ${TEST_OPTS}
+	  				 -e DEIS_ROUTER_SERVICE_PORT=${DEIS_ROUTER_SERVICE_PORT} \
+						 -e TEST_OPTS=${TEST_OPTS}
+						 -e DEFAULT_EVENTUALLY_TIMEOUT="30s" ${IMAGE}
```
arschles: nit: do these spacings line up in an editor?

jchauncey: yeah not sure wtf github is doing here

tests/admin_test.go
```diff
@@ -0,0 +1,94 @@
...
+			sess, err := start("deis auth:regenerate -u %s", adminTestData.Profile, testData.Username)
+			Eventually(sess).Should(Say("Token Regenerated"))
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		// It("regenerates the token for all users", func() {
+		// 	sess, err := start("deis auth:regenerate --all", adminTestData.Profile)
+		// 	Eventually(sess).Should(Say("Token Regenerated"))
+		// 	Expect(err).NotTo(HaveOccurred())
+		// })
```
arschles: Can you remove this commented code or mark pending (i.e. `XIt`)? In either case, probably a good idea to put in an issue if not there already.

jchauncey: k i marked it pending and added a comment as to why it was marked that way. I'll also open an issue

tests/builds_test.go
```diff
@@ -53,25 +55,27 @@ var _ = Describe("Builds", func() {
...
 			})
 
 			It("can list app builds", func() {
-				cmd, err := start("deis builds:list -a %s", testApp.Name)
+				sess, err := start("deis builds:list -a %s", testData.Profile, testApp.Name)
+				Eventually(sess).Should(Say(uuidRegExp))
+				Eventually(sess).Should(Exit(0))
 				Expect(err).NotTo(HaveOccurred())
-				Eventually(cmd).Should(Exit(0))
-				Eventually(cmd).Should(Say(uuidRegExp))
```
arschles: Why switch the order of these checks? Did parallel execution require this change? This type of change seems to be repeated throughout this PR.

tests/builds_test.go
```diff
@@ -104,39 +110,39 @@ var _ = Describe("Builds", func() {
...
 				Expect(len(scrapeProcs(testApp.Name, procsListing))).To(Equal(2))
 
 				// TODO: https://github.com/deis/workflow-e2e/issues/84
 				// "deis logs -a %s", app
-				// sess, err = start("deis logs -a %s", testApp.Name)
+				// sess, err = start("deis logs -a %s", testData.Profile, testApp.Name)
 				// Expect(err).To(BeNil())
 				// Eventually(sess).Should(Say("hi"))
-				// Eventually(sess).Should(Exit(0))
+				// Eventually(sess).Should(Exit(0))                                           Expect(err).NotTo(HaveOccurred())
```
arschles: Can you put the `Expect` on a new line?

tests/certs_test.go
```diff
@@ -26,76 +25,78 @@ type Cert struct {
...
 	Eventually(sess).Should(Exit(0))
+	Expect(err).NotTo(HaveOccurred())
 
-	Eventually(listCerts().Wait().Out.Contents()).Should(ContainSubstring(certName))
+	Eventually(listCerts(profile).Wait().Out.Contents()).Should(ContainSubstring(certName))
 }
 
-func attachCert(certName, domain string) {
-	attachOrDetachCert(certName, domain, "attach")
+func attachCert(profile string, certName, domain string) {
```
arschles: you can remove the `string` after `profile`

tests/certs_test.go
```diff
@@ -26,76 +25,78 @@ type Cert struct {
...
 	}
 
-	Eventually(listCerts()).Should(Say("No certs"))
+	Eventually(listCerts(profile)).Should(Say("No certs"))
 }
 
-func addCert(certName, cert, key string) {
-	sess, err := start("deis certs:add %s %s %s", certName, cert, key)
-	Expect(err).NotTo(HaveOccurred())
+func addCert(profile string, certName, cert, key string) {
```
arschles: you can remove the `string` after `profile`

tests/certs_test.go
```diff
@@ -26,76 +25,78 @@ type Cert struct {
...
 
-func attachCert(certName, domain string) {
-	attachOrDetachCert(certName, domain, "attach")
+func attachCert(profile string, certName, domain string) {
+	attachOrDetachCert(profile, certName, domain, "attach")
 }
 
-func detachCert(certName, domain string) {
-	attachOrDetachCert(certName, domain, "detach")
+func detachCert(profile string, certName, domain string) {
```
arschles: same here and below regarding the `string` after `profile`

tests/builds_test.go
jchauncey: so i noticed that when we were waiting on things to be returned from the CLI if we had statements like check for the exit code those worked better if they were at the end. It may be completely coincidence just an observation so i tried to make all of them consistent. On the other hand I also tried to check for errors first `Expect(err)`.

arschles: Ah, ok. Maybe related to waiting for the buffer to fill up. Thanks for explaining, and sgtm

tests/healthcheck_test.go
```diff
@@ -11,29 +11,21 @@ import (
...
-			sess, err := start("deis apps:destroy --confirm=%s", appName)
-			Expect(err).To(BeNil())
 			Eventually(sess).Should(Exit(0))
-			Eventually(sess).Should(Say("Destroying %s...", appName))
-			Eventually(sess).Should(Say("done in "))
-			Eventually(sess).Should(Say("Git remote deis removed"))
+			Expect(err).NotTo(HaveOccurred())
 		})
 
 		// TODO: test is broken
```
arschles: is there an issue for fixing this? out of scope for this PR and not blocking though

tests/limits_test.go
```diff
@@ -13,62 +15,66 @@ var _ = Describe("Limits", func() {
...
+		var testData TestData
 
 		BeforeEach(func() {
-			// Set up the Limits test app only once and assume the suite will clean up.
-			once.Do(func() {
-				os.Chdir("example-go")
-				appName := getRandAppName()
-				createApp(appName)
-				testApp = deployApp(appName)
-			})
```
arschles: we can remove the one-time execution b/c each `It` is run in parallel, right?

arschles: Let me amend my previous comment: we can remove because each `It` is run in a separate process...

tests/tests_suite_test.go
arschles: @vdice I think it's here: https://github.com/jchauncey/workflow-e2e/blob/random-user/tests/tests_suite_test.go#L115

vdice: got it, thanks @arschles !

vdice:
@jchauncey note the TravisCI build is failing as it can't find the `build` make target (commented out in this PR)

jchauncey:
Doh over site sorry. I remove that because we no longer build the binary. I was just using docker-build. Will probably need to link that to build

Makefile
```diff
@@ -24,15 +25,19 @@ bootstrap:
 
 .PHONY: test-integration
 test-integration:
-	go test ./tests/... ${TEST_OPTS}
+	DEFAULT_EVENTUALLY_TIMEOUT="30s" ginkgo $TEST_OPTS tests/
```
vdice: The default in master for sequential tests is `10s` and it has proven to be _relatively_ consistent with that value.  Can we keep it at 10?

Makefile
```diff
@@ -8,7 +8,8 @@ DEV_IMG := quay.io/deis/go-dev:0.9.0
 RUN_CMD := docker run --rm -e DEIS_ROUTER_SERVICE_HOST=${DEIS_ROUTER_SERVICE_HOST} -e DEIS_ROUTER_SERVICE_PORT=${DEIS_ROUTER_SERVICE_PORT} -v ${CURDIR}:${SRC_PATH} -w ${SRC_PATH} ${DEV_IMG}
 DEV_CMD := docker run --rm -e GO15VENDOREXPERIMENT=1 -v ${CURDIR}:${SRC_PATH} -w ${SRC_PATH} ${DEV_IMG}
 
-TEST_OPTS := -test.v -test.timeout=60m -ginkgo.v -ginkgo.slowSpecThreshold=120
+TEST_OPTS := -slowSpecThreshold=120.00 -noisyPendings=false
+PARALLEL_TEST_OPTS := -slowSpecThreshold=120.00 -noisyPendings=false -p
```
vdice: Looks like this can be simplified to:
```
TEST_OPTS := -slowSpecThreshold=120.00 -noisyPendings=false
PARALLEL_TEST_OPTS := ${TEST_OPTS} -p
```

Makefile
jchauncey: so ive made enough changes to the tests that I had significant failures with a 10s timeout. I am weary to move it that low given the failures i was seeing.

tests/limits_test.go
jchauncey: yes we dont need this. We want each test (It block) to receive its own test data

tests/healthcheck_test.go
jchauncey: for fixing what?

tests/admin_test.go
arschles: :+1:

tests/healthcheck_test.go
arschles: The broken test? Or, is that an old comment?

tests/limits_test.go
arschles: :+1: sgtm

tests/healthcheck_test.go
jchauncey: thats an old comment

Issue #139, opened 2016-03-30
Question: do we need the time.Sleep in BeforeSuite?


arschles:
The comment at https://github.com/deis/workflow-e2e/blob/master/tests/tests_suite_test.go#L130 says `// wait for ssh key to propagate. `, but I don't think we need that anymore since the builder no longer relies on confd to propagate SSH keys. It instead checks key validity from the controller on-demand.

Can we remove that sleep?

helgi:
I say remove the sleep, if it doesn't work without it then I'd consider it a bug

PR #143, opened 2016-03-31
fix(tests_suite_test.go): remove sleep

arschles:
… that was being used to wait for SSH key propagation (which no longer
should happen)

Fixes https://github.com/deis/workflow-e2e/issues/139


jchauncey:
this needs rebasing

arschles:
@jchauncey rebased

PR #164, opened 2016-04-13
fix(ps_test): remove ps:list check for 'restart "one"` type

vdice:


jchauncey:
🚢

krancour:
Taking away one lgtm... I think we should consider @helgi's suggestion of leaving these enabled, but disabling some assertions against `ps:list` after `ps:restart` that may have been in excess.

jchauncey:
If `ps:list` is reporting the status of a pod as `up` when its really `terminating` then that should be fixed. Until we can fix that bug I say we mark the tests as pending, open an issue for fixing `ps:list` and make that the highest priority item to fix.

bacongobbler:
> Until we can fix that bug I say we mark the tests as pending, open an issue for fixing ps:list and make that the highest priority item to fix.

agreed.

krancour:
@jchauncey there's no disputing that it needs to be fixed.  I think the issue here is that is doesn't make sense to disable tests for functionality that is actually working (`ps:restart` actually works fine, it would seem) if we can avoid it.  The way to avoid it is by dropping a couple assertions that were probably going above and beyond what needed to be asserted in that spec.

I think there's room for another story in `ps_test.go` that's specifically focused on ensuring the accuracy of the `ps:list` results.  That doesn't exist yet, but if it did, _that_ is the one that should be pending.

krancour:
We're back to two lgtms, but before we do this, let's seriously consider that we're disabling tests for functionality that actually works.  I'd so much rather we keep those to ward of regressions.

vdice:
Testing locally w/ the following amendment, all else same as origin/master:
```
@@ -147,7 +147,9 @@ var _ = Describe("deis ps", func() {
                                        afterProcs := scrapeProcs(app.Name, procsListing)

                                        // compare the before and after sets of process names
-                                       Expect(len(afterProcs)).To(Equal(scaleTo))
+                                       if restart != "one" {
+                                               Expect(len(afterProcs)).To(Equal(scaleTo))
+                                       }
                                        if scaleTo > 0 && restart != "by wrong type" {
                                                Expect(beforeProcs).NotTo(Equal(afterProcs))
                                        }
```
Is this along the lines of what you were thinking, @krancour ?

krancour:
@vdice #166 is what I had in mind.

bacongobbler:
> #166 is what I had in mind.

That looks a little destructive, considering that it removes all functionality of checking the processes restarted are up for every test case. I think @vdice's approach is much gentler, though a comment saying that it's a workaround for a bug in deis/controller is probably in order.

krancour:
@bacongobbler question is if `ps:restart` already shows you the new pods names that have replaced the old, what point is there at all in running `ps:list`?

But if we prefer @vdice's approach, I am not opposed to it.

bacongobbler:
I see your point, but this code caught that the controller was only returning 3 out of 4 pods after a restart, which tells me there's a bug in the controller. I don't think it's excessive to think that after restarting a process, you should expect to see the same number of pods in `deis ps:list`.

tests/ps_test.go
```diff
@@ -147,7 +147,9 @@ var _ = Describe("deis ps", func() {
 					afterProcs := scrapeProcs(app.Name, procsListing)
 
 					// compare the before and after sets of process names
-					Expect(len(afterProcs)).To(Equal(scaleTo))
+					if restart != "one" { // current workaround for an issue in deis/controller
```
mboersma: Is there an issue to track that we want to restore this test check in the future? And I'd add a `TODO:` or `HACK:` comment prefix to underline that idea.

Also, e2e has passed without this change, so I wonder if it is truly needed now.

jchauncey: That is the only test that reliably failed for me when testing his PR
locally

bacongobbler: I've filed https://github.com/deis/controller/issues/631 to track this behaviour.

vdice: Thanks @bacongobbler

bacongobbler: @mboersma workflow-e2e build 649 passed with this change, but the rest of the job list is a bunch of red: https://ci.deis.io/job/workflow-test-pr/

PR #166, opened 2016-04-13
chore(ps_test): Remove excessive assertions that cause test failure

krancour:
Alternative to #164 

Instead of marking these tests pending, this removes some assertions that were probably excessive to begin with, thus allowing us to keep these tests.


jchauncey:
running these tests now

jchauncey:
did not see a failure from the `ps_test`. I had other failures so rerunning again.

jchauncey:
I'm confident that this will fix out `ps:test` problem for now.

bacongobbler:
this failed for unrelated reasons, but #164 was merged which fixes the same root issue this PR was trying to solve. Shall we close this one out and track the root issue at deis/controller#631?

krancour:
Closing this because #164 was merged.

PR #171, opened 2016-04-14
feat(git_push_test): curl app endpoint for "Powered by"

### LINKED ISSUE #170: Tests should curl a buildpack app
Apparently there isn't a full-roundtrip `git push` test involving a buildpack app, so it's possible for builder to return success (mistakenly) but have the app actually fail to deploy, and e2e won't notice.

Which actually happened--true story! :book:

mboersma:
Closes #170.

### DIFF
--- a/tests/git_push_test.go
+++ b/tests/git_push_test.go
@@ -56,7 +56,7 @@ var _ = Describe("git push deis master", func() {
 					})
 
 					Specify("that user can deploy that app using a git push", func() {
-						git.Push(user, keyPath)
+						git.Push(user, keyPath, app)
 					})
 
 				})
--- a/tests/cmd/git/commands.go
+++ b/tests/cmd/git/commands.go
@@ -1,6 +1,10 @@
 package git
 
 import (
+	"fmt"
+	"net/http"
+	"strconv"
+
 	"github.com/deis/workflow-e2e/tests/cmd"
 	"github.com/deis/workflow-e2e/tests/model"
 	"github.com/deis/workflow-e2e/tests/settings"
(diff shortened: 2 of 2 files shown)


tests/git_push_test.go
```diff
@@ -57,6 +60,13 @@ var _ = Describe("git push deis master", func() {
 
 					Specify("that user can deploy that app using a git push", func() {
 						git.Push(user, keyPath)
+						// curl the app's root URL and print just the HTTP response code
+						cmdRetryTimeout := 60
+						curlCmd := model.Cmd{CommandLineString: fmt.Sprintf(`curl -sL -w "%%{http_code}\\n" "%s" -o /dev/null`, app.URL)}
+						Eventually(cmd.Retry(curlCmd, strconv.Itoa(http.StatusOK), cmdRetryTimeout)).Should(BeTrue())
+						// verify that the response contains "Powered by" as all the example apps do
+						curlCmd = model.Cmd{CommandLineString: fmt.Sprintf(`curl -sL "%s"`, app.URL)}
+						Eventually(cmd.Retry(curlCmd, "Powered by ", cmdRetryTimeout)).Should(BeTrue())
```
krancour: So, I think this is the only `git push` that happens anywhere in the e2e tests currently, but in the interest of keeping the tests orderly as me move forward, here's a question...

Wouldn't we _always_ want to curl after a `git push` in order to assert the success of that deployment?  If so, I think those assertions should be built right into (moved to) `git.Push(...)` in the `tests/cmd/git` package.  My intention had been that under `cmd`, we amass lots of useful bit like this that we repeats often, and that would even include assertions like these wherever possible... all with the goal of keeping the specs themselves super concise and readable.

Just food for thought.

mboersma:
Jenkins, test this please.

mboersma:
This seems to be failing because this is the first test spec to deploy a Buildpack app, and therefore it pays the price for faulting in slugbuilder/slugrunner and times out. That could be worked around by "priming" the cluster with those images, but really `git push deis master` shouldn't be returning until the app is routable.

helgi:
We're lacking timing information in there but I suspect it goes beyond the `timeout` limit we have set around deploys and that's why it is returning.

We may have to set higher limits for `build pack` apps *or* detect somehow the image is being downloaded still and give it a longer timeout when the download takes a long time

mboersma:
Several fixes to deis/controller recently have made this test stable. :anchor: 

I'm merging this and then creating another PR with a similar spec for a Dockerfile-based app, since that has a different deployment path on the back end.

PR #177, opened 2016-04-21
feat(registry): add tests for private registry functionality

helgi:
This depends on the following PRs:

* https://github.com/deis/jenkins-jobs/pull/63
* https://github.com/deis/controller/pull/653
* https://github.com/deis/workflow-cli/pull/45

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@ RUN apk add --update-cache \
 	&& go get -u -v \
 	github.com/tools/godep \
 	github.com/onsi/ginkgo/ginkgo \
-	&& curl -L https://dl.bintray.com/deis/deisci/deis-6c176d2-linux-amd64 -o /usr/local/bin/deis \
+	&& curl -L https://dl.bintray.com/deis/deisci/deis-852b0b0-linux-amd64 -o /usr/local/bin/deis \
 	&& chmod +x /usr/local/bin/deis \
 	&& mkdir -p $GOPATH/src/k8s.io \
 	&& curl -L https://github.com/kubernetes/kubernetes/archive/v$K8S_VERSION.tar.gz | tar xvz -C $GOPATH/src/k8s.io \
--- a/tests/registry_test.go
+++ b/tests/registry_test.go
@@ -1,0 +1,166 @@
+package tests
+
+import (
+	"time"
+
+	"github.com/deis/workflow-e2e/tests/cmd"
+	"github.com/deis/workflow-e2e/tests/cmd/apps"
+	"github.com/deis/workflow-e2e/tests/cmd/auth"
+	"github.com/deis/workflow-e2e/tests/model"
+	"github.com/deis/workflow-e2e/tests/settings"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/extensions/table"
+	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gbytes"
+	. "github.com/onsi/gomega/gexec"
+)
+
+var _ = Describe("deis registry", func() {
+
+	Context("with an existing user", func() {
+
+		var user model.User
+
+		BeforeEach(func() {
+			user = auth.Register()
+		})
+
+		AfterEach(func() {
+			auth.Cancel(user)
+		})
+
+		Context("who owns an existing app", func() {
+
+			var app model.App
+
+			BeforeEach(func() {
+				app = apps.Create(user, "--no-remote")
+			})
+
+			AfterEach(func() {
+				apps.Destroy(user, app)
+			})
+
+			Specify("that user can list that app's registry information", func() {
+				sess, err := cmd.Start("deis registry:list --app=%s", &user, app.Name)
+				Eventually(sess).Should(Say("=== %s Registry", app.Name))
+				Expect(err).NotTo(HaveOccurred())
+				Eventually(sess).Should(Exit(0))
+			})
+
+			Specify("that user cannot unset an invalid registry information", func() {
+				sess, err := cmd.Start("deis registry:unset --app=%s munkafolyamat", &user, app.Name)
+				// TODO: should unsetting a bogus registry info return 0 (success?)
+				Eventually(sess, settings.MaxEventuallyTimeout).Should(Say("=== %s Registry", app.Name))
+				Eventually(sess).ShouldNot(Say(`munkafolyamat\s+yeah`, app.Name))
+				Expect(err).NotTo(HaveOccurred())
+				Eventually(sess).Should(Exit(0))
+			})
+
+			Specify("that user can set a valid registry information", func() {
+				sess, err := cmd.Start("deis registry:set --app=%s username=bob", &user, app.Name)
+				Eventually(sess, settings.MaxEventuallyTimeout).Should(Say("=== %s Registry", app.Name))
+				Eventually(sess).Should(Say(`username\s+bob`))
+				Expect(err).NotTo(HaveOccurred())
+				Eventually(sess).Should(Exit(0))
+
+				sess, err = cmd.Start("deis registry:list --app=%s", &user, app.Name)
+				Eventually(sess).Should(Say("=== %s Registry", app.Name))
+				Eventually(sess).Should(Say(`username\s+bob`))
+				Expect(err).NotTo(HaveOccurred())
+				Eventually(sess).Should(Exit(0))
+			})
+
+			Specify("that user can not deploy from a private registry due to lack of credentials", func() {
+				// do an unsuccessful deploy
+				image := "quay.io/deisci/e2e-private-registry-test"
+				sess, err := cmd.Start("deis pull --app=%s %s", &user, app.Name, image)
+				Expect(err).NotTo(HaveOccurred())
+				Eventually(sess).Should(Say("Creating build..."))
+				Eventually(sess, settings.MaxEventuallyTimeout).Should(Exit(1))
+				time.Sleep(10 * time.Second)
+			})
+
+			Specify("that user can deploy from a private registry using registry credentials", func() {
+				// read-only access
+				registry_creds := "TP5BS3NHW0OZ20GER4IORTIJF90J48KKJ8NX8YC7Z22N5P7WE27BRKVMQ4QAEID8"
+				sess, err := cmd.Start("deis registry:set --app=%s username=deisci+e2e_registry password=%s", &user, app.Name, registry_creds)
+				Eventually(sess, settings.MaxEventuallyTimeout).Should(Say("=== %s Registry", app.Name))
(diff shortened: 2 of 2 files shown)


tests/cmd/builds/commands.go
```diff
@@ -19,16 +19,20 @@ const ExampleImage = "deis/example-dockerfile-http"
...
 // Pull executes the `deis pull` shortcut as the specified user.
-func Pull(user model.User, app model.App) {
-	createOrPull(user, app, "pull")
+func Pull(user model.User, app model.App, image string) {
+	createOrPull(user, app, "pull", image)
 }
 
-func createOrPull(user model.User, app model.App, command string) {
-	sess, err := cmd.Start("deis %s --app=%s %s", &user, command, app.Name, ExampleImage)
+func createOrPull(user model.User, app model.App, command string, image string) {
```
krancour: imo, this could be implemented more cleanly using variadic arguments.

Method signature:

```
func createOrPull(user model.User, app model.App, command string, opts ...string)
```
`Create(...)` and `Pull(...)` functions can have their signatures modified similarly.

By doing this, invocations of the `Create(...)` or `Pull(...)` functions will not require the awkward `""` at the tail end of the argument list.  Specifying an image would be fully optional.  `createOrPull(...)` can retrieve it from `opts[0]`.

tests/registry_test.go
```diff
@@ -0,0 +1,167 @@
...
+
+				BeforeEach(func() {
+					sess, err := cmd.Start("deis registry:set --app=%s username=bob", &user, app.Name)
+					Eventually(sess, settings.MaxEventuallyTimeout).Should(Say("=== %s Registry", app.Name))
+					Eventually(sess).Should(Say(`username\s+bob`))
+					Expect(err).NotTo(HaveOccurred())
+					Eventually(sess).Should(Exit(0))
+				})
+
+				Specify("that user can unset that tag from that app", func() {
```
krancour: tag?

tests/registry_test.go
```diff
@@ -0,0 +1,167 @@
...
+
+		BeforeEach(func() {
+			user = auth.Register()
+		})
+
+		AfterEach(func() {
+			auth.Cancel(user)
+		})
+
+		Context("who owns an existing app that has already been deployed", func() {
```
krancour: The "that has already been deployed" bit is actually inconsequential to most of the specs within this context.  For instance, the fact that the app has already been deployed is irrelevant when testing a user's ability or inability to add or modify registry credentials.  By eliminating that, you can eliminate unnecessary pulls that create a drag on the e2e tests.

tests/registry_test.go
```diff
@@ -0,0 +1,167 @@
...
+					Eventually(sess).Should(Say("=== %s Registry", app.Name))
+					Eventually(sess).ShouldNot(Say(`username\s+bob`))
+					Eventually(sess).ShouldNot(Say(`munkafolyamat\s+yeah`, app.Name))
+					Expect(err).NotTo(HaveOccurred())
+					Eventually(sess).Should(Exit(0))
+				})
+
+			})
+
+			Context("and an application is deployed using a private registry image", func() {
```
krancour: Usually there would be a "before" to satisfy the conditions described by the context.

tests/registry_test.go
helgi: For the pure set / unset / list parts I needed an existing app. I could shove that into the `Successful` spec potentially

tests/registry_test.go
helgi: I'm not sure I understand what you talking about

tests/registry_test.go
krancour: Yes... you need an existing apo, but it doesn't need to have been deployed already, right?

helgi: hmm, yeah that's technically true. I don't particularly care if it has a deploy. I'll modify

tests/registry_test.go
krancour: Context establishes a precondition that has been satisfied before each spec within runs.  You have context here that doesn't actually do anything before each spec to satisfy the condition it describes.

helgi: That's a good point, I originally had a BeforeEach but that's obviously gone now. Removing

tests/cmd/builds/commands.go
```diff
@@ -18,17 +18,22 @@ import (
...
-func Pull(user model.User, app model.App) {
-	createOrPull(user, app, "pull")
+func Pull(user model.User, app model.App, opts ...string) {
+	createOrPull(user, app, "pull", opts)
 }
 
-func createOrPull(user model.User, app model.App, command string) {
-	sess, err := cmd.Start("deis %s --app=%s %s", &user, command, app.Name, ExampleImage)
+func createOrPull(user model.User, app model.App, command string, opts []string) {
+	image := opts[0]
```
bacongobbler: I'm not quite sure this is a good way to send an option list to `createOrPull`. How about a `struct` instead?

```
type CreateOrPullOpts struct {
    Image string
}
```

That way we can directly reference them instead of having to keep track of where certain values are stored in a slice.

tests/registry_test.go
```diff
@@ -0,0 +1,162 @@
...
+				image := "quay.io/deisci/e2e-private-registry-test"
+				sess, err := cmd.Start("deis pull --app=%s %s", &user, app.Name, image)
+				Expect(err).NotTo(HaveOccurred())
+				Eventually(sess).Should(Say("Creating build..."))
+				Eventually(sess, settings.MaxEventuallyTimeout).Should(Exit(1))
+				time.Sleep(10 * time.Second)
+			})
+
+			Specify("that user can deploy with registry credentials", func() {
+				sess, err := cmd.Start("deis registry:set --app=%s username=deisci+e2e_registry password=%s", &user, app.Name, os.Getenv("QUAY_ROBOT_ACCOUNT_TOKEN"))
```
bacongobbler: this username/password should be public knowledge or else we won't be able to run e2e anywhere but in jenkins.

tests/cmd/builds/commands.go
helgi: It was on a suggestion from @krancour - `struct` just seemed like an overkill but I guess is one way Go does function defaults

tests/registry_test.go
helgi: It is both a junk account (read-only) and a junk registry

bacongobbler: right, but I don't see anywhere that $QUAY_ROBOT_ACCOUNT_TOKEN is set, so if I run the tests locally it will fail. Where do I retrieve this token?

helgi: That is in the Jenkins secret storage, but that's a fair point, I'm adding a real token here since it is fairly innocent access, read-only to a repo with only a renamed `example-go` in it

tests/cmd/builds/commands.go
krancour: I think a struct is overkill as well.  I have encountered many, many examples of variadic arguments being used in this fashion.  I'll concede that just because I've frequently seen it done does not implicitly make it right.  But I don't see the harm in it either.  This is all largely a matter of opinion.  Ultimately, I am fine with wherever we land on this _if_ the end result is that invocations of the method are clear and concise.  In a test suite like this, I would consider clarity and conciseness paramount.  Each spec tells a story and weird, extraneous details (like needing to create a struct that encapsulates some options) takes you out of that story.

tests/registry_test.go
```diff
@@ -0,0 +1,166 @@
...
+				Eventually(sess).Should(Exit(0))
+
+				sess, err = cmd.Start("deis registry:list --app=%s", &user, app.Name)
+				Eventually(sess).Should(Say("=== %s Registry", app.Name))
+				Eventually(sess).Should(Say(`username\s+bob`))
+				Expect(err).NotTo(HaveOccurred())
+				Eventually(sess).Should(Exit(0))
+			})
+
+			Specify("that user can not deploy due to lack of credentials", func() {
```
krancour: May I suggest: "that user can not deploy _from a private registry_ due to lack of credentials"?

tests/registry_test.go
```diff
@@ -0,0 +1,166 @@
...
+				// do an unsuccessful deploy
+				image := "quay.io/deisci/e2e-private-registry-test"
+				sess, err := cmd.Start("deis pull --app=%s %s", &user, app.Name, image)
+				Expect(err).NotTo(HaveOccurred())
+				Eventually(sess).Should(Say("Creating build..."))
+				Eventually(sess, settings.MaxEventuallyTimeout).Should(Exit(1))
+				time.Sleep(10 * time.Second)
+			})
+
+			Specify("that user can deploy with registry credentials", func() {
```
krancour: Similarly to above, may I suggest: "that user can deploy _from a private registry_ using registry credentials"

krancour:
Apart from a couple suggestions on rewording some spec titles (nits, really), this LGTM.  These are very nicely written.  👍

Dockerfile
```diff
@@ -12,7 +12,7 @@ RUN apk add --update-cache \
 	&& go get -u -v \
 	github.com/tools/godep \
 	github.com/onsi/ginkgo/ginkgo \
-	&& curl -L https://dl.bintray.com/deis/deisci/deis-6c176d2-linux-amd64 -o /usr/local/bin/deis \
+	&& curl -L https://dl.bintray.com/deis/deisci/deis-852b0b0-linux-amd64 -o /usr/local/bin/deis \
```
mboersma: Is there a way just to point to the "latest" cli built from master? It seems tedious to update this SHA continually, although I know that's not in the scope of this PR.

tests/registry_test.go
```diff
@@ -0,0 +1,166 @@
...
+			Specify("that user can list that app's registry information", func() {
+				sess, err := cmd.Start("deis registry:list --app=%s", &user, app.Name)
+				Eventually(sess).Should(Say("=== %s Registry", app.Name))
+				Expect(err).NotTo(HaveOccurred())
+				Eventually(sess).Should(Exit(0))
+			})
+
+			Specify("that user cannot unset an invalid registry information", func() {
+				sess, err := cmd.Start("deis registry:unset --app=%s munkafolyamat", &user, app.Name)
+				// TODO: should unsetting a bogus registry info return 0 (success?)
```
mboersma: My vote is that it should return non-zero, although there may be other `deis` commands that don't in similar cases.

Dockerfile
helgi: I had a quick look at the time and couldn't find it so I punted

tests/registry_test.go
helgi: That's a copy from the tags test, actually :) So blame whoever wrote the tags test ;)

Dockerfile
krancour: I would say that fixing that is beyond the scope of this PR.  Although we should still figure that out.

PR #180, opened 2016-04-26
chore(logs): Add deis logs test

jchauncey:


tests/apps_test.go
```diff
@@ -177,12 +177,11 @@ var _ = Describe("deis apps", func() {
 
 			// Pending: The logging subsystem is not currently installed as part of the workflow-dev
 			// chart.
```
krancour: Should remove this comment.

tests/apps_test.go
```diff
@@ -175,14 +175,11 @@ var _ = Describe("deis apps", func() {
 				Eventually(sess).Should(Exit(0))
 			})
 
-			// Pending: The logging subsystem is not currently installed as part of the workflow-dev
-			// chart.
-			XSpecify("that user can retrieve logs for that app", func() {
+			Specify("that user can retrieve logs for that app", func() {
 				sess, err := cmd.Start("deis logs -a %s", &user, app.Name)
```
arschles: not too familiar with this `Start` func, so ignore me if I'm mistaken, but is the first parameter supposed to be a format string? if so, there are not enough format directives (i.e. `%s`) for the number of params given

jchauncey: `start` takes the user running the command as the second param.

arschles: Nevermind, after talking privately I understand that this runs the command as the given used, with `app.Name` as the single param to the format string

PR #184, opened 2016-04-29
feat(deploy): test that setting DEPLOY_BATCHES on an app works

helgi:
https://github.com/deis/controller/pull/665

### DIFF
--- a/tests/builds_test.go
+++ b/tests/builds_test.go
@@ -6,6 +6,7 @@ import (
 	"github.com/deis/workflow-e2e/tests/cmd/auth"
 	"github.com/deis/workflow-e2e/tests/cmd/builds"
 	"github.com/deis/workflow-e2e/tests/model"
+	"github.com/deis/workflow-e2e/tests/settings"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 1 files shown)


krancour:
This is a good thing to test.  👍 

If we want to keep this super clean, this new spec should actually moved down, under the existing context "who owns an existing app that has already been deployed".  Another nested context should probably be added under that-- something like "and that app has been scaled out to 11 instances".  These changes would cause the contexts to very clearly define the preconditions for the test case.

My only other concern about this... where / how can we assert that new deployments actually roll containers 5 at a time?  The assertion that are in place now only verify that that config was set, which means this actually doesn't differ much from existing `deis config:set` specs.

krancour:
And just for tracking purposes... this looks like another pesky case of CI having been re-triggered (and incorrectly so) by a comment.

helgi:
We can't really assert it beside looking at the controller logs - maybe we can constantly review the pods via kubectl

krancour:
That seems complicated.  Alternatively, would there be any way of updating controller and/or the cli to actually indicate how many pods get rolled at a time right before the `..o` wait for deployment starts?  That could give us something to assert upon that wasn't there before.

helgi:
That's. Pretty fundamental change in the API and CLI that requiring streaming or polling, we have support for neither

jchauncey:
ci failed @krancour

jchauncey:
@helgi can we move this to rc1?

PR #190, opened 2016-05-10
chore(apps:create): Assert on new/improved CLI output

krancour:
Complements deis/workflow-cli#57


krancour:
CI on this will fail until deis/workflow-cli#57 is merged.

PR #210, opened 2016-05-16
test(auth_test.go): add spec for changing password

### LINKED ISSUE #195: No password change tests
`deis auth:passwd` allows a user to change their password, but isn't covered in e2e tests.

mboersma:
Closes #195.


Joshua-Anderson:
Should we also have some test coverage for an admin changing another user's password?

krancour:
@Joshua-Anderson, this is nice, though I have a suggestion. This spec relies on the user already being logged in as a precondition. What I'd typically recommend is that be handled using a context block wrapped around this spec.  The `auth.Login()` on the first line of your spec could move there, _but_ since the `auth.Register()` further up implicitly logs the user in, you could probably just put a comment in the new context to explain that it doesn't do anything, but that it exists to add clarity and improve the narrative (specs are stories) for its constituent specs.

Also, there are some existing specs in here that reallly, logically have that same precondition that the user must be logged in. examples would be specs for `whoami` and regenerating one's own token. It's _my_ oversight that that context was missed, but if you were to improve your spec by adding new context as I have suggested, you could also move some of the other specs, as appropriate, into that new context.

mboersma:
> admin changing another user's password?

Definitely, I'll add that.

> be handled using a context block wrapped around this spec

I'll take a look at that refactoring as well.

vdice:
ping @mboersma on the status of this PR -- still scheduled for the `v2.0-rc1` milestone?

bacongobbler:
pinging @mboersma

mboersma:
> ready for review or still in progress?

It's still in progress, because I intend to implement @krancour's excellent suggestions but probably won't be able to in time for the v2.0 release. Since these are e2e improvements, they don't depend strictly on the release cycle anyway, so I'll remove the milestone.

Joshua-Anderson:
@mboersma Are you still interested in e2e tests that change a users password? There aren't tests for it yet in the CLI but there will be some soon 😄

Since this is tested in the SDK, CLI, and controller, it probably doesn't need an e2e test as well.

mboersma:
> Are you still interested in e2e tests that change a users password?

Probably not if we're covered on those other fronts. We want to prefer local unit tests in general rather than put the burden on e2e. I'll revisit this and see if it's still useful and either fix or close it.

mboersma:
This ancient PR is covered in other tests well enough IMHO, so I'm closing it.

PR #212, opened 2016-05-17
test(apps_test): add more specs for "deis run"

### LINKED ISSUE #204: No complex command tests for "deis run"
The existing tests mostly run `env`, but there should be an example of a command that itself contains dashes and flags, so the `--` style syntax can be validated:
```console
$ deis run -a myapp -- ls -alh && echo "-- dashes --"
```

mboersma:
This covers some commands that caused problems in the past.

~~This also necessarily includes the change from #211. I will rebase it after that merges.~~ Rebased.

Requires deis/controller#751.

Closes #204.

### DIFF
--- a/tests/apps_test.go
+++ b/tests/apps_test.go
@@ -190,9 +190,41 @@ var _ = Describe("deis apps", func() {
 
 			Specify("that user can run a command in that app's environment", func() {
 				sess, err := cmd.Start("deis apps:run --app=%s echo Hello, 世界", &user, app.Name)
-				Eventually(sess, settings.MaxEventuallyTimeout).Should(Say("Hello, 世界"))
 				Expect(err).NotTo(HaveOccurred())
+				Eventually(sess, (settings.MaxEventuallyTimeout)).Should(Say("Hello, 世界"))
 				Eventually(sess).Should(Exit(0))
+			})
+
+			Specify("that user can run a command with dashes in that app's environment", func() {
+				sess, err := cmd.Start("deis apps:run --app=%s -- ls -alh", &user, app.Name)
+				Expect(err).NotTo(HaveOccurred())
+				// Can't assume too much about arbitrary "ls" output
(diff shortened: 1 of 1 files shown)


helgi:
Should we throw in a `run` calling a fake binary to test a failure scenario?

mboersma:
@helgi I added a test for a non-existent command--is that what you had in mind?

But it uncovered that `deis apps:run /usr/bin/boguscmd` returns 0 and prints `No such file or directory` to stdout, not stderr. Those seem like bugs to me so I'll enter them at workflow-cli.

helgi:
@mboersma yes sir!

As far as the bug, could also be related to the new way we do deis run in the API

helgi:
Oh nevermind, I see why CLI

tests/apps_test.go
```diff
@@ -190,11 +190,43 @@ var _ = Describe("deis apps", func() {
...
+				sess, err := cmd.Start("deis apps:run --app=%s dd if=/dev/urandom bs=3072 count=1000", &user, app.Name)
+				Expect(err).NotTo(HaveOccurred())
+				Eventually(sess, (settings.MaxEventuallyTimeout)).Should(Exit(0))
+				Expect(len(sess.Out.Contents())).To(BeNumerically(">=", 3072000))
+			})
+
+			Specify("that user can't run a bogus command in that app's environment", func() {
+				sess, err := cmd.Start("deis apps:run --app=%s /usr/bin/boguscmd", &user, app.Name)
+				Expect(err).NotTo(HaveOccurred())
+				// TODO: should this have gone to stderr instead?
```
vdice: have you answered your question here @mboersma ?

vdice:
Added a comment re: the `TODO` note but not blocking for this PR; code LGTM.  (Do note that the jenkins job failure needs investigation...)

tests/apps_test.go
mboersma: See deis/workflow-cli#69.

PR #217, opened 2016-06-01
chore(Dockerfile): update Workflow CLI to latest published

mboersma:
Closes #216.

### DIFF
--- a/tests/apps_test.go
+++ b/tests/apps_test.go
@@ -204,7 +204,8 @@ var _ = Describe("deis apps", func() {
 				Eventually(sess).Should(Exit(0))
 			})
 
-			Specify("that user can run a command with quotes in that app's environment", func() {
+			// FIXME: why is Workflow failing this test?
+			PSpecify("that user can run a command with quotes in that app's environment", func() {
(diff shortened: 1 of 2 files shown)


mboersma:
I also added a commit that marks one Ginkgo test spec as pending, because it seems to fail often with the current `deis` CLI (for me locally as well).

tests/apps_test.go
```diff
@@ -204,7 +204,8 @@ var _ = Describe("deis apps", func() {
 				Eventually(sess).Should(Exit(0))
 			})
 
-			Specify("that user can run a command with quotes in that app's environment", func() {
+			// FIXME: why is Workflow failing this test?
```
arschles: is there an issue for this?

arschles:
LGTM3, although an issue referenced in https://github.com/deis/workflow-e2e/pull/217#discussion_r65396318 would be helpful

tests/apps_test.go
mboersma: Assuming this PR passes--that the `FIXME:` is needed--I'll create an issue to track it before I merge this.

PR #219, opened 2016-06-03
fix(apps_test): escape quotes properly in deis run spec

### LINKED ISSUE #218: Fix "deis run with quotes" test spec
See this [`// FIXME:` comment](https://github.com/deis/workflow-e2e/blob/master/tests/apps_test.go#L207) from #217.

mboersma:
Closes #218.


bacongobbler:
punting this to 2.1

PR #226, opened 2016-06-09
fix(registry): recorganise registry tests order and fix a should(say()) call

helgi:


arschles:
@helgi code LGTM, although e2e failed :(

PR #230, opened 2016-06-15
feat(git_push_test.go): add git push interrupt test

### LINKED ISSUE #207: Add test to ensure git pushes can be interrupted and subsequent ones will pass
After https://github.com/deis/builder/pull/339, a `git push` operation against builder should be able to get killed, and subsequent ones will then succeed (instead of failing with "another push is in progress")

vdice:
[skip ci] Depends on (currently pending) fix for https://github.com/deis/builder/issues/365

Closes #207


vdice:
Closing this PR as the issue that spawned it (https://github.com/deis/workflow-e2e/issues/207) has been closed.

Issue #167, opened 2016-04-13
Test suite should build deis CLI from source


mboersma:
Currently changes to workflow-cli won't be tested fully in e2e since the [client is downloaded from bintray](https://github.com/deis/workflow-e2e/blob/0b0156d020e3eec1f9bc586818b3286f97dde3e4/Dockerfile#L9).

Joshua-Anderson:
Would it maybe make more sense to download the client binary ***of the commit being tested*** from bintray so that the binary that is being distributed to people is the one being tested?

bacongobbler:
Is there a link to retrieve a specific commit? That would also be handy for `install-v2.sh` because that script only grabs the latest client (which a user may not want when installing 2.0 a few months down the road, for example)

Joshua-Anderson:
> Is there a link on bintray to retrieve a specific commit?

`https://bintray.com/deis/deisci/download_file?file_path=deis-<first 7 char of git sha>-<os>-<arch>` seems to work.

mboersma:
See #217. Still updating the test binary manually by SHA. 😞

sgoings:
What about if we wrote a lil' shell script to check to see if WORKFLOW_CLI_SHA (or something like that) is in the Docker run environment, and if so do what @Joshua-Anderson suggests. If not, download the latest version. Essentially it would be a deis cli shim...

arschles:
note that now we've moved built binaries from bintray to google cloud storage, a sample URL to retrieve the CLI for a specific SHA would be: https://storage.googleapis.com/workflow-cli/08a0e40/deis-08a0e40-linux-amd64

Joshua-Anderson:
I opened PR #233 to fix this.

PR #231, opened 2016-06-16
feat(apps): apps:run will now print on stderr if the command errors

Joshua-Anderson:
deis/workflow-cli#81
Requires workflow-cli#7d1fe129bc6721c31940a5d93a5d396a4d5839df


Joshua-Anderson:
This is failing because the CLI is being cached by docker, but I'm not sure what to do about it.

Joshua-Anderson:
As a temporary fix I'm just changing docker to download the version of the CLI this PR requires. We'll talk more about the underlying cause in #167

PR #233, opened 2016-06-17
feat(docker): install specified CLI version at runtime

### LINKED ISSUE #167: Test suite should build deis CLI from source

Joshua-Anderson:
This downloads the CLI builds from gcs when the docker container is run. It defaults to latest but can be overridden if necessary to use a specific git sha.

Fixes #167


Joshua-Anderson:
This fixes #167

Issue #224, opened 2016-06-07
Add test for concurrent pushes to the builder


arschles:
Two types of tests should be added:

- Concurrent pushes to different repos (all should succeed)
- Concurrent pushes to the same repo (only 1 should succeed)

mboersma:
Test specs that cover this were added to the e2e tests in deis/workflow#235, but this references builder unit or functional tests as well. So let's leave it open but not consider it done yet.

mboersma:
@arschles, is this closed now after deis/builder#361? There are both flavors of concurrent push tests now in e2e, and at least the second case (same repo) in the builder func tests now.

arschles:
@mboersma https://github.com/deis/builder/pull/361 addresses this issue at a unit-test level. I believe that's enough coverage for this particular test case. I'd like to get at least your thoughts as well, though - and anyone else who you believe should weigh in if you think it's appropriate.

mboersma:
> I believe that's enough coverage for this particular test case.

Agreed. I'll go ahead and close this, but if anyone thinks there are more nuances we should capture here, please reopen this or create a new ticket.

PR #235, opened 2016-06-17
test(git_push): add concurrent push specs

mboersma:
```console
$ FOCUS_TEST="git push deis master" DEIS_CONTROLLER_URL=your.workflow.host make test-integration
...
Ran 6 of 179 Specs in 364.039 seconds
SUCCESS! -- 6 Passed | 0 Failed | 0 Pending | 173 Skipped 
```
(Add `-v` in the Makefile `TEST_OPTS` to see hilariously interleaved buildpack/dockerfile output.)

Refs #224.

cc: @vdice.

TODO:
- [x] `curl` app endpoints in different app specs
- [x] loop until success in same app specs
- [x] use `filepath` for portability

### DIFF
--- a/tests/cmd/git/commands.go
+++ b/tests/cmd/git/commands.go
@@ -25,11 +25,16 @@ const (
 
 // Push executes a `git push deis master` from the current directory using the provided key.
 func Push(user model.User, keyPath string, app model.App, banner string) {
-	sess := doGitPush(user, keyPath)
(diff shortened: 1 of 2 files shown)


vdice:
@mboersma 
> loop until success in same app specs

Is covered in https://github.com/deis/workflow-e2e/pull/237 if we want to pull that logic in...

PR #237, opened 2016-06-20
feat(git_push_test.go): add git push interrupt test

vdice:
~~Still some TODOs to address that I can move forward with if it sounds like this will still be wanted after~~ ~~Still need to align with work done in https://github.com/deis/workflow-e2e/pull/235~~

Adds spec for https://github.com/deis/builder/pull/369

cc @mboersma

### DIFF
--- a/tests/cmd/git/commands.go
+++ b/tests/cmd/git/commands.go
@@ -3,23 +3,29 @@ package git
 import (
 	"fmt"
 	"net/http"
+	"os"
 	"strconv"
+	"time"
 
 	"github.com/deis/workflow-e2e/tests/cmd"
 	"github.com/deis/workflow-e2e/tests/model"
 	"github.com/deis/workflow-e2e/tests/settings"
 
 	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gbytes"
 	. "github.com/onsi/gomega/gexec"
 )
 
 // The functions in this file implement SUCCESS CASES for commonly used `git` commands.
 // This allows each of these to be re-used easily in multiple contexts.
 
+const (
+	pushCommandLineString = "GIT_SSH=%s GIT_KEY=%s git push deis master"
+)
+
 // Push executes a `git push deis master` from the current directory using the provided key.
 func Push(user model.User, keyPath string, app model.App, banner string) {
-	sess, err := cmd.Start("GIT_SSH=%s GIT_KEY=%s git push deis master", &user, settings.GitSSH, keyPath)
-	Expect(err).NotTo(HaveOccurred())
+	sess := doGitPush(user, keyPath)
 	// sess.Wait(settings.MaxEventuallyTimeout)
 	// output := string(sess.Out.Contents())
 	// Expect(output).To(MatchRegexp(`Done, %s:v\d deployed to Deis`, app.Name))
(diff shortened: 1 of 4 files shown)


vdice:
OK to test

tests/cmd/helper.go
```diff
@@ -66,6 +66,8 @@ func StartCmd(command model.Cmd) (*gexec.Session, error) {
 // supplied <timeout> until the <cmd> result contains the <expectedResult>
 // An example use of this utility would be curl-ing a url and waiting
 // until the response code matches the expected response.
+// TODO: refactor to accept timeout of type time.Duration, potentially expected exit code was well
+// TODO: or at least use RetryUntilResult below
```
arschles: @vdice can we get some GH issues for these, if they're not already in?

tests/cmd/git/commands.go
```diff
@@ -33,3 +39,36 @@ func Push(user model.User, keyPath string, app model.App, banner string) {
...
 	Eventually(cmd.Retry(curlCmd, banner, cmdRetryTimeout)).Should(BeTrue())
 }
+
+// PushWithInterrupt executes a `git push deis master` from the current
+// directory using the provided key, but then halts the progress via SIGINT.
+func PushWithInterrupt(user model.User, keyPath string) {
+	sess := doGitPush(user, keyPath)
+	Eventually(sess.Err).Should(Say("Starting build... but first, coffee!"))
+
+	sess = sess.Interrupt()
```
arschles: this is a bummer b/c it could introduce a timing dependency (the git push could complete before this executes)

I know there's no reasonable fix right now, so this is me just recording the thought. consider it the equivalent of the tiniest nit for now!

tests/cmd/helper.go
```diff
@@ -81,3 +83,40 @@ func Retry(command model.Cmd, expectedResult string, timeout int) bool {
...
+
+	fmt.Fprintf(ginkgo.GinkgoWriter,
+		"Waiting up to %d seconds for `%s` to return expected cmdResult %v...\n",
+		int(timeout.Seconds()), command.CommandLineString, expectedCmdResult)
+
+	tck := time.NewTicker(period)
+	tmr := time.NewTimer(timeout)
+	defer tck.Stop()
+	defer tmr.Stop()
+	totalTicks := 0
```
arschles: I see that this var is incremented below, but where is its value used?

tests/git_push_test.go
```diff
@@ -59,6 +59,17 @@ var _ = Describe("git push deis master", func() {
 						git.Push(user, keyPath, app, "Powered by Deis")
 					})
 
+					Specify("that user can interrupt the deploy of the app and re-convene", func() {
```
arschles: maybe say "recover" instead of "re-convene"?

tests/cmd/git/commands.go
```diff
@@ -33,3 +39,36 @@ func Push(user model.User, keyPath string, app model.App, banner string) {
...
+// directory using the provided key, until the command result satisfies
+// expectedCmdResult of type model.CmdResult, failing if
+// settings.DefaultEventuallyTimeout is reached first.
+func PushUntilResult(user model.User, keyPath string, expectedCmdResult model.CmdResult) {
+	envVars := append(os.Environ(), fmt.Sprintf("DEIS_PROFILE=%s", user.Username))
+	pushCmd := model.Cmd{Env: envVars, CommandLineString: fmt.Sprintf(
+		pushCommandLineString, settings.GitSSH, keyPath)}
+
+	Eventually(cmd.RetryUntilResult(pushCmd, expectedCmdResult, 5*time.Second,
+		settings.DefaultEventuallyTimeout)).Should(BeTrue())
```
arschles: can the default eventually timeout be specified in the env? wondering because we may need to change the builder's timeout sometime in the future

tests/model/model.go
```diff
@@ -72,3 +73,26 @@ func getDir() string {
...
+// ExitCode
+type CmdResult struct {
+	Out      []byte
+	Err      []byte
+	ExitCode int
+}
+
+// Satisfies returns whether or not the original CmdResult, ocd, meets all of
+// the expectations contained in the expeced CmdResult, ecd
+func (ocd CmdResult) Satisfies(ecd CmdResult) bool {
```
arschles: this receiver name describes my coding style perfectly

tests/cmd/git/commands.go
vdice: @arschles yes, it appears it and `MaxDefaultEventuallyTimeout` can be overridden via env vars: https://github.com/deis/workflow-e2e/blob/master/tests/settings/settings.go#L20

tests/cmd/git/commands.go
vdice: @arschles Yes, good point.  Although gomega's [Eventually](https://onsi.github.io/gomega/#making-asynchronous-assertions) polls the `sess` provided and returns immediately once the supplied string is found (or, if never found, hits the default timeout), it seems anytime 'Eventually' is used on a `sess`, there's no way to actually interrupt/kill/stop that session.  I just noticed when I turned on verbose output, the original `git push deis master` cmd is still ongoing in the background.

Alternatively, if we pull out the `Eventually` check, the cmd is indeed terminated... but then we'd end up needing an even uglier `time.Sleep(duration)` to wait for the cmd to progress a bit before interrupting.  Thoughts?

tests/cmd/helper.go
vdice: Thanks @arschles , it wasn't needed/used.  Removed.

tests/git_push_test.go
vdice: done!

tests/cmd/git/commands.go
arschles: I think what's here is fine for now

PR #238, opened 2016-06-20
chore(Makefile): remove deprecated '-f' flag

vdice:


vdice:
actual 'test' of this change passed in the build job where tagging is done: https://ci.deis.io/job/workflow-e2e-pr/338/

PR #239, opened 2016-06-22
feat(*): update to use new errors from CLI

Joshua-Anderson:
Note that this will not pass until deis/workflow-cli#92 has been merged.


tests/certs_test.go
```diff
@@ -43,7 +43,7 @@ var _ = Describe("deis certs", func() {
 
 		Specify("that user cannot add a cert with a malformed name", func() {
 			sess, err := cmd.Start("deis certs:add %s %s %s", &user, "bogus.cert.name", cert.CertPath, cert.KeyPath)
-			Eventually(sess.Err).Should(Say("400 Bad Request"))
+			Eventually(sess.Err).Should(Say("Can only contain a-z (lowercase), 0-9 and hypens"))
```
vdice: s/hypens/hyphens

Joshua-Anderson: Hah. Just noticed that. It's actually a controller issue... the typo is in there.

Joshua-Anderson: https://github.com/deis/controller/blob/master/rootfs/api/models/__init__.py#L34

tests/apps_test.go
```diff
@@ -51,7 +51,7 @@ var _ = Describe("deis apps", func() {
 
 			Specify("that user cannot get information about that app", func() {
 				sess, err := cmd.Start("deis info -a %s", &user, bogusAppName)
-				Eventually(sess.Err).Should(Say("Not found."))
+				Eventually(sess.Err).Should(Say("Error: Not Found"))
```
vdice: Understanding this PR just followed the precedent before of repeated/duplicated expected cli errors, wondering what your thoughts are on creating re-usable vars (perhaps in https://github.com/deis/workflow-e2e/blob/master/tests/settings/settings.go or maybe elsewhere) for the most commonly expected error outputs.  

For instance, `"Error: Not Found"` could be the value for `var ErrNotFound` and then it'd just need to be changed in one spot when/if the cli error changes.  Just a thought and def not blocking for this PR...

tests/certs_test.go
vdice: ah so it is!

tests/apps_test.go
Joshua-Anderson: Even better, it could use errors from the go SDK! Good idea.

PR #248, opened 2016-06-29
fix(.): clean up docker-test-integration shell script

Joshua-Anderson:
This doesn't fix the issue where `echo "Installing Workflow CLI version $CLI_VERSION"` is not properly printing the version, but it does clean up the script a bit.

### DIFF
--- a/docker-test-integration.sh
+++ b/docker-test-integration.sh
@@ -6,11 +6,11 @@
 BASE_URL="https://storage.googleapis.com/workflow-cli"
 URL="$BASE_URL/deis-latest-linux-amd64"
 
-if [[ $CLI_VERSION -ne "latest" ]]; then
+if [[ "$CLI_VERSION" != "latest" ]]; then
 	URL="$BASE_URL/$CLI_VERSION/deis-$CLI_VERSION-linux-amd64"
 fi
 
 echo "Installing Workflow CLI version $CLI_VERSION"
-curl $URL -o /usr/local/bin/deis && chmod +x /usr/local/bin/deis
+curl "$URL" -o /usr/local/bin/deis && chmod +x /usr/local/bin/deis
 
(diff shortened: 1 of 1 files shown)


Joshua-Anderson:
For some reason, though this script works for me locally, it fails on jenkins. Not sure what is going on.

bacongobbler:
Probably because `!=` is not a valid equality statement in bash (have to use `ne`).

Is this PR outdated now? Looks like it just encapsulates a variable in quotes and changes the syntax from `-ne` to `!=`.

jchauncey:
@bacongobbler 

```
!=
is not equal to

if [ "$a" != "$b" ]

This operator uses pattern matching within a [[ ... ]] construct.
```

from - http://tldp.org/LDP/abs/html/comparison-ops.html

bacongobbler:
Hmm, I guess I'm mixing it up with /bin/sh again...

Manual testing:

```
><> if [[ "a" != "b" ]]; then echo hi; fi
hi
><> if [[ "a" -ne "b" ]]; then echo hi; fi
><>
```

it's interesting that the first evaluation succeeds, but the second doesn't. Anyone have an idea why?

Either way I guess this has an LGTM from me

bacongobbler:
Oh. If I actually read the documentation:

> Note that integer and string comparison use a different set of operators.

vdice:
@Joshua-Anderson I didn't see that this fix was already PR-ed when I issued a similar in https://github.com/deis/workflow-e2e/pull/260.  Let me know if I should close 260 -- though also note there is a bug in jenkins-jobs that https://github.com/deis/jenkins-jobs/pull/147 proposes to fix that is related as well.

Joshua-Anderson:
@vdice I like your PR better, as it fixes more things, so let's go with that instead.

PR #251, opened 2016-07-11
feat(deployments): split out tests for deployment specific tests

helgi:
Should be eventually removed

Requires deis/controller#854

Adds a `config:set` cmd that allows us to set `DEIS_KUBERNETES_DEPLOYMENTS=1` as needed

### DIFF
--- a/tests/cmd/configs/commands.go
+++ b/tests/cmd/configs/commands.go
@@ -1,0 +1,24 @@
+package configs
+
+import (
+	"github.com/deis/workflow-e2e/tests/cmd"
+	"github.com/deis/workflow-e2e/tests/model"
+	"github.com/deis/workflow-e2e/tests/settings"
+
+	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gbytes"
+	. "github.com/onsi/gomega/gexec"
+)
+
+// The functions in this file implement SUCCESS CASES for commonly used `deis config` subcommands.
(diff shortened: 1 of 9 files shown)


tests/cmd/configs/commands.go
```diff
@@ -0,0 +1,24 @@
...
+// The functions in this file implement SUCCESS CASES for commonly used `deis config` subcommands.
+// This allows each of these to be re-used easily in multiple contexts.
+
+// Set executes `deis config:set` on the specified app as the specified user.
+func Set(user model.User, app model.App, key string, value string) *Session {
+	sess, err := cmd.Start("deis config:set %s=%s --app=%s", &user, key, value, app.Name)
+	Expect(err).NotTo(HaveOccurred())
+	sess.Wait(settings.MaxEventuallyTimeout)
+	Eventually(sess).Should(Say("Creating config..."))
+	Eventually(sess).Should(Exit(0))
```
arschles: If these matchers fail, they'll always report the same line number, which will be a bummer for debugging. Possible to move them outside the function (in a reusable manner)? If it's not possible, not a dealbreaker.

helgi: This is how we do all other things in `cmd` tho - Would you suggest changing all of that?

arschles: nope, if it's already how we do things, then I'm good with leaving it :+1:

PR #252, opened 2016-07-11
feat(git): ensure deis run works for git push

helgi:


tests/git_push_test.go
```diff
@@ -128,6 +128,13 @@ var _ = Describe("git push deis master", func() {
 
 					})
 
+					Specify("and can execute deis run successfully", func() {
+						git.Push(user, keyPath, app, "Powered by Deis")
+						sess, err := cmd.Start("deis run env -a %s", &user, app.Name)
+						Expect(err).NotTo(HaveOccurred())
+						Eventually(sess).Should(Exit(0))
```
arschles: is it worth capturing the output of the command here and in the below test?

arschles:
@helgi looks good, even if you don't end up checking the output in https://github.com/deis/workflow-e2e/pull/252#discussion-diff-70704700

tests/git_push_test.go
helgi: Not really, we don't care about the output, at all, just that it runs successfully

PR #253, opened 2016-07-12
chore(Makefile): use latest go-dev and glide

mboersma:
~~In particular, updates controller-sdk-go to the current tip of master.~~

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,7 @@ export DEIS_CONTROLLER_URL=http://deis.${DEIS_ROUTER_SERVICE_HOST}.nip.io
 endif
 endif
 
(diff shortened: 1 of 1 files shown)


Joshua-Anderson:
@mboersma How is this related to updating controller-sdk-go in e2e?

krancour:
This LGTM, but I have the same confusion as @Joshua-Anderson. I think the description in the OP might have been copy/pasted and doesn't match the title and the diffs (which at least agree with each other).

mboersma:
> How is this related to updating controller-sdk-go in e2e?

It isn't, any longer. I had issues recreating the glide.yaml / glide.lock successfully so just tossed out that commit for now. :-/ Sorry for the confusion.

Issue #208, opened 2016-05-13
No "git push" test for Dockerfile with Procfile


bacongobbler:
`git push deis master` with a Dockerfile and a Procfile is currently untested.

test steps:

 - create an app like example-dockerfile-http with a committed dockerfile and procfile with a `web` process type defined
 - git push deis master
 - observe that the deployed process is of type "web"

bacongobbler:
created https://github.com/deis/example-dockerfile-procfile-http for this test case

PR #257, opened 2016-07-13
feat(procfile): add support to test dockerfile app with proctype WEB

smothiki:
fixes https://github.com/deis/workflow-e2e/issues/208

### DIFF
--- a/tests/deployments_dockerfiles_test.go
+++ b/tests/deployments_dockerfiles_test.go
@@ -40,7 +40,7 @@ var _ = Describe("all dockerfile apps", func() {
 			})
 
 			DescribeTable("can deploy an example dockerfile app",
-				func(url, buildpack, banner string) {
+				func(url, buildpack, banner, proctype string) {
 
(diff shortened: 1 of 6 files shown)


mboersma:
This looks great @smothiki. My only concern is that these test specs are part of the default `SKIP_TEST` regex, so they aren't run in e2e, just manually on-demand. (We still need to implement a nightly job for "all buildpacks|dockerfiles.")

I think covering the dockerfile + buildpack case should be included in CI runs for each PR, because controller changes could break this complicated behavior. Maybe we could move this test to a different go file, or tweak the regex so that this is run by default?

smothiki:
I'm leaving the changes in the dockerfile tests as it is. Modified git push tests

mboersma:
> Modified git push tests

That should cover it until we figure out how/when we want to run the skipped specs. Thanks!

helgi:
Remember to update the `deployment_` duplicates with the same changes

PR #261, opened 2016-07-15
feat(buildpacks): fetch unique consecutive buildpacks each time

smothiki:
Get unique consecutive build packs for each build

### DIFF
--- a/tests/git_push_test.go
+++ b/tests/git_push_test.go
@@ -12,6 +12,7 @@ import (
 	"github.com/deis/workflow-e2e/tests/cmd/keys"
 	"github.com/deis/workflow-e2e/tests/model"
 	"github.com/deis/workflow-e2e/tests/settings"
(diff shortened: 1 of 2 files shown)


mboersma:
Testing that all buildpacks function is important, but I don't think e2e should cycle through them on consecutive test runs. Some of them are faster and more reliable than others (for example, the go buildpack will retry if temporary network problems prevent it from downloading the Go tarball, but scala just quits if you look at it wrong).

We had also considered picking a buildpack at random for each run, but decided against it in Deis v1. The argument is/was that tests should be as deterministic as possible.

I'd still prefer to solve the "all buildpacks" problem by creating a nightly job that runs the tests we already have, or by just biting the bullet and not skipping them any longer despite the increased test run time. But this is just my opinion and I'm curious what others think.

smothiki:
Anyways this PR is based on build_number I don't think BUILD_NUMBER is set as env variable. SO closing it

PR #263, opened 2016-07-18
feat(apptypes): change integration script to test buildpack and docke…

smothiki:
requires a change in charts and jenkins-jobs repo to actually test this behavior


smothiki:
have been restarting this job since yesterday. Tests are failing for other reasons not from this PR. Merging the PR as this PR only added a if clause to choose which kind of job to run. As expected the e2e jobs are running as bps env var is not set

PR #271, opened 2016-07-22
feat(shortcuts): add shortcuts test

### LINKED ISSUE #243: add e2e tests for `deis shortcuts`
see https://github.com/deis/workflow-cli/pull/96

sgoings:
Fixes #243


sgoings:
Feel free to tell me how my code is noob-y. 😆

helgi:
Should this not be in the CLI instead of part of e2e?

sgoings:
Ha! @helgi that's a good point, and probably would lead to faster fixes if there ever were problems. I'll move it over.

PR #272, opened 2016-07-24
fix(certs): limit=0 is invalid in DRF 3.4

helgi:
https://github.com/tomchristie/django-rest-framework/pull/4194

Requires deis/controller#892

### DIFF
--- a/tests/certs_test.go
+++ b/tests/certs_test.go
@@ -180,8 +180,11 @@ var _ = Describe("deis certs", func() {
 			Specify("that user can limit the number of certs returned by certs:list", func() {
 				randCertRegExp := `\d{0,9}-cert`
 
-				sess, err := cmd.Start("deis certs:list -l 0", &user)
-				Eventually(sess).Should(Say("No certs"))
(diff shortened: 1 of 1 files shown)


tests/certs_test.go
```diff
@@ -180,8 +180,11 @@ var _ = Describe("deis certs", func() {
 			Specify("that user can limit the number of certs returned by certs:list", func() {
 				randCertRegExp := `\d{0,9}-cert`
 
-				sess, err := cmd.Start("deis certs:list -l 0", &user)
-				Eventually(sess).Should(Say("No certs"))
+				// limit=0 is invalid as of DRF 3.4
+				// https://github.com/tomchristie/django-rest-framework/pull/4194
+				sess, err := cmd.Start("deis certs:list --limit=0", &user)
+				Eventually(sess).Should(Say(randCertRegExp))
+				Eventually(sess).Should(Say(randCertRegExp))
```
kmala: this is same as the line above....is it intended?

helgi: yeah, because it is replacing `no cert` with matching 2 certs

PR #273, opened 2016-07-25
feat(controller): deis pull failure should check if the existing rc/pods are not deleted

kmala:


tests/builds_test.go
```diff
@@ -73,6 +73,7 @@ var _ = Describe("deis builds", func() {
 			})
 
 			Specify("that user can't create a new build of that app from a nonexistent image using `deis pull`", func() {
+				builds.Create(user, app)
```
mboersma: We already created the app with `apps.Create()` in the `BeforeEach()`, so this shouldn't be needed.

PR #298, opened 2016-08-08
chore(travis): Remove .travis.yml

jchauncey:


vdice:
https://github.com/deis/workflow-e2e/blob/master/_scripts/deploy.sh can go as well.

PR #305, opened 2016-08-10
chore(glide.lock): update ginkgo, gomega, controller-sdk-go

mboersma:
Updates docker-go-dev and three core e2e vendored packages to their latest commits.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,3 @@
-export GO15VENDOREXPERIMENT=1
-
 SHORT_NAME := workflow-e2e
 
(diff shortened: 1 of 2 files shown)


bacongobbler:
is this in progress or ready for review?

mboersma:
@bacongobbler this is still in progress--it seems to be failing for valid reasons (the test container is never seen as "started"). Tests run fine locally, and I haven't been able to pinpoint the difference when they're in the workflow-e2e Docker container yet.

PR #308, opened 2016-08-12
feat(maintenance): Add test for maintenance mode

kmala:
ref https://github.com/deis/workflow-cli/pull/171

### DIFF
--- a/tests/maintenance_test.go
+++ b/tests/maintenance_test.go
@@ -1,0 +1,94 @@
+package tests
+
+import (
+	"fmt"
+	"strconv"
+
+	"github.com/deis/workflow-e2e/tests/cmd"
+	"github.com/deis/workflow-e2e/tests/cmd/apps"
+	"github.com/deis/workflow-e2e/tests/cmd/auth"
+	"github.com/deis/workflow-e2e/tests/cmd/builds"
+	"github.com/deis/workflow-e2e/tests/model"
+	"github.com/deis/workflow-e2e/tests/settings"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gbytes"
+	. "github.com/onsi/gomega/gexec"
+)
+
+var _ = Describe("deis maintenance", func() {
+
+	Context("with an existing user", func() {
+
+		var user model.User
+
+		BeforeEach(func() {
+			user = auth.Register()
+		})
+
+		AfterEach(func() {
+			auth.Cancel(user)
+		})
+
+		Context("who owns an existing app that has already been deployed", func() {
+
+			var app model.App
+
+			BeforeEach(func() {
+				app = apps.Create(user, "--no-remote")
+				builds.Create(user, app)
+			})
+
+			AfterEach(func() {
+				apps.Destroy(user, app)
+			})
+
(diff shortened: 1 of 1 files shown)


tests/maintenance_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+		BeforeEach(func() {
+			user = auth.Register()
+		})
+
+		AfterEach(func() {
+			auth.Cancel(user)
+		})
+
+		Context("who owns an existing app that has already been deployed", func() {
```
krancour: "that has already been deployed" isn't relevant to the spec titled "that user can list that app's maintenance info." In general, avoiding unnecessary deployments helps the speed of the test suite overall.

krancour:
One line note. Otherwise LGTM.

mboersma:
This still needs some attention, so I'm moving it to the next release cycle.

tests/maintenance_test.go
kmala: changed the spec.

tests/maintenance_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+				sess, err = cmd.Start("deis maintenance:info --app=%s", &user, app.Name)
+				Eventually(sess).Should(Say("Maintenance mode is on.\n"))
+				Expect(err).NotTo(HaveOccurred())
+				Eventually(sess).Should(Exit(0))
+
+				// curl the app's root URL and print just the HTTP response code
+				cmdRetryTimeout := 60
+				curlCmd := model.Cmd{CommandLineString: fmt.Sprintf(`curl -sL -w "%%{http_code}\\n" "%s" -o /dev/null`, app.URL)}
+				Eventually(cmd.Retry(curlCmd, strconv.Itoa(503), cmdRetryTimeout)).Should(BeTrue())
```
mboersma: Should this spec finish with `deis maintenance:off`? I don't see that command being tested anywhere.

kmala: added the tests

tests/maintenance_test.go
```diff
@@ -0,0 +1,128 @@
...
+				Eventually(sess).Should(Exit(0))
+
+				cmdRetryTimeout = 60
+				curlCmd = model.Cmd{CommandLineString: fmt.Sprintf(`curl -sL -w "%%{http_code}\\n" "%s" -o /dev/null`, app.URL)}
+				Eventually(cmd.Retry(curlCmd, strconv.Itoa(200), cmdRetryTimeout)).Should(BeTrue())
+			})
+		})
+	})
+
+	DescribeTable("any user can get command-line help for maintenance", func(command string, expected string) {
```
Joshua-Anderson: Please don't add a section for help.

These will be covered in CLI tests and make e2e very brittle.

bacongobbler: every other command currently has e2e tests for this, though I agree we can just leave this to cli/sdk unit tests instead of e2e to make e2e more stable/faster.

Joshua-Anderson: > every other command currently has e2e tests for this

As soon as I get code coverage in this area in the CLI I hope to remove them 😄

kmala: i had added because we are testing it currently for all other commands.I am ok to remove this if cli/sdk unit tests are going to take care of it.

PR #310, opened 2016-08-13
fix(Makefile): add retrying to workflow CLI download

helgi:
Also run chmod as a separate command as conditionals negate exit on error that -e is trying to provide
This should prevent the CI system form continuing when there is a problem downloading the CLI and also should let GCS settle down a bit after the binary is uploaded by providing the retry and some preprive (10 seconds) between each retry


helgi:
Added response header dumping so that we can compare our custom headers (manually) to what the binary thinks it is. Will help with debuging

helgi:
Things are not green since an older binary is still being downloaded but we have enough data going on now to pin point things better so I am going to merge this

PR #313, opened 2016-08-18
ref(Dockerfile): don't use alpine, it has known DNS issues

Joshua-Anderson:
cc @jchauncey @helgi


helgi:
What is the golang image based on?

bacongobbler:
Alpine. Check the tag ;)

> FROM golang:1.6-alpine

Joshua-Anderson:
Nope @helgi: `golang:1.7`, debian jessie

bacongobbler:
if you're referring to go-dev: https://github.com/deis/docker-go-dev/blob/master/rootfs/Dockerfile

bacongobbler:
@Joshua-Anderson sorry, I thought @helgi was referring to the old golang image, not go-dev.

Joshua-Anderson:
This change seems to have broken `git remote` tests in some mysterious way (exit code 129). However, they are going to be implemented in the cli very soon regardless. Is it okay to remove these tests for now?

This did uncover that I am not properly logging the `STDERR` from git when weird errors like this come up, so I plan to fix that as well.

PR #315, opened 2016-08-22
feat(docker-test-integration.sh): try curling cli from multiple buckets

vdice:
Now that we have separate `-pr` and `-master` buckets that could feasibly
contain a given cli binary.

Ref https://github.com/deis/workflow-cli/pull/190

### DIFF
--- a/docker-test-integration.sh
+++ b/docker-test-integration.sh
@@ -19,17 +19,23 @@ function debug {
 
 trap debug ERR
 
-BASE_URL="https://storage.googleapis.com/workflow-cli"
-URL="${BASE_URL}/deis-latest-linux-amd64"
+curl-cli-from-gcs-bucket() {
+	local gcs_bucket="${1}"
+	local base_url="https://storage.googleapis.com/${gcs_bucket}"
+	local url="${base_url}/deis-latest-linux-amd64"
 
-if [[ "${CLI_VERSION}" != "latest" ]]; then
-	URL="${BASE_URL}/${CLI_VERSION}/deis-${CLI_VERSION}-linux-amd64"
-fi
+	if [[ "${CLI_VERSION}" != "latest" ]]; then
+		url="${base_url}/${CLI_VERSION}/deis-${CLI_VERSION}-linux-amd64"
+	fi
 
-# Download CLI, retry up to 5 times with 10 second delay between each
-echo "Installing Workflow CLI version '${CLI_VERSION}' via url '${URL}'"
-curl --silent --show-error -I "${URL}"
-curl --silent --show-error --retry 5 --retry-delay 10 -o /usr/local/bin/deis "${URL}"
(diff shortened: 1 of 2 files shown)


docker-test-integration.sh
```diff
@@ -19,17 +19,23 @@ function debug {
...
-if [[ "${CLI_VERSION}" != "latest" ]]; then
-	URL="${BASE_URL}/${CLI_VERSION}/deis-${CLI_VERSION}-linux-amd64"
-fi
+	if [[ "${CLI_VERSION}" != "latest" ]]; then
+		url="${base_url}/${CLI_VERSION}/deis-${CLI_VERSION}-linux-amd64"
+	fi
+
+	# Download CLI, retry up to 5 times with 10 second delay between each
+	echo "Installing Workflow CLI version '${CLI_VERSION}' via url '${url}'"
+	curl -f --silent --show-error -I "${url}"
```
vdice: I added the `-f/--fail` flag back in so we could get non-zero exit code from curl if the http response code was >= 400 (see also https://curl.haxx.se/mail/archive-2006-02/0004.html).

docker-test-integration.sh
```diff
@@ -19,17 +19,23 @@ function debug {
...
+	curl -f --silent --show-error --retry 5 --retry-delay 10 -o /usr/local/bin/deis "${url}"
+}
 
-# Download CLI, retry up to 5 times with 10 second delay between each
-echo "Installing Workflow CLI version '${CLI_VERSION}' via url '${URL}'"
-curl --silent --show-error -I "${URL}"
-curl --silent --show-error --retry 5 --retry-delay 10 -o /usr/local/bin/deis "${URL}"
+# try both buckets for specific CLI_VERSION
+curl-cli-from-gcs-bucket "workflow-cli-master" || curl-cli-from-gcs-bucket "workflow-cli-pr"
```
Joshua-Anderson: Temporarily, we might else want to try `workflow-cli` as well to avoid breakage.

vdice: @Joshua-Anderson good point -- I bet we don't have a `latest` binary in either bucket yet.  I'll do so... could you write up the ticket to remove use of the old `workflow-cli` bucket?  Perhaps in `v2.6` or sooner?

vdice: @Joshua-Anderson on second thought, I don't think we should support curling from `workflow-cli` unless there are PRs in flight from before https://github.com/deis/workflow-cli/pull/181 that we need to run e2e on (even then, I'd suggest rebasing on top of the changes in 181).  Although we could remove them, I wouldn't want ci running with a `latest` binary from `workflow-cli` when there exists one in `workflow-cli-master` (which one does as of this writing)  Thoughts?

Joshua-Anderson: Yeah, thinking about it this wouldn't be too useful, I didn't realize how few PRs this would affect. Disregard!

PR #317, opened 2016-08-25
feat(whitelist): Add tests for deis whitelist

kmala:


tests/whitelist_test.go
```diff
@@ -0,0 +1,93 @@
...
+				// curl the app's root URL and print just the HTTP response code
+				cmdRetryTimeout := 60
+				curlCmd := model.Cmd{CommandLineString: fmt.Sprintf(`curl -sL -w "%%{http_code}\\n" "%s" -o /dev/null`, app.URL)}
+				Eventually(cmd.Retry(curlCmd, strconv.Itoa(200), cmdRetryTimeout)).Should(BeTrue())
+			})
+
+			Specify("can add/remove addresses from the whitelist", func() {
+				sess, err := cmd.Start("deis whitelist:add 1.2.3.4 --app=%s", &user, app.Name)
+				Eventually(sess, settings.MaxEventuallyTimeout).Should(Say("done"))
+				Expect(err).NotTo(HaveOccurred())
```
mboersma: In general, I think the `Expect(err).NotTo(HaveOccurred())` should be the first check since it's just looking directly at the `err` returned from `cmd.Start`. If that returned an error, it's probably not worth following the `Eventually` checks that use goroutines and timers to examine `stdout` text.

Just a nitpick--the test is valid either way.

PR #320, opened 2016-09-02
fix(healthcheck): use correct port the image exposes

kmala:
with this PR https://github.com/deis/controller/pull/1036 the healthchecks are properly checked and hence the e2e fails.

### DIFF
--- a/tests/healthcheck_test.go
+++ b/tests/healthcheck_test.go
@@ -65,34 +65,33 @@ var _ = Describe("deis healthchecks", func() {
 				Eventually(sess).Should(Exit(0))
 			})
 
-			// disable httpGet liveness checks as it won't pass for every app
-			// TODO(bacongobbler): somehow determine *what* port we need to set the liveness check for this app
-			XSpecify("that user can set a httpGet liveness healthcheck", func() {
-				sess, err := cmd.Start("deis healthchecks:set liveness httpGet -a %s 80", &user, app.Name)
+			// 1500 is the port of the app we are deploying deis/example-dockerfile-http
+			Specify("that user can set a httpGet liveness healthcheck", func() {
+				sess, err := cmd.Start("deis healthchecks:set liveness httpGet -a %s 1500", &user, app.Name)
 				Eventually(sess).Should(Say("Applying livenessProbe healthcheck..."))
 				Eventually(sess, settings.MaxEventuallyTimeout).Should(Say("=== %s Healthchecks", app.Name))
-				Eventually(sess).Should(Say(`HTTP GET Probe\: Path="/" Port=80 HTTPHeaders=\[]`))
+				Eventually(sess).Should(Say(`HTTP GET Probe\: Path="/" Port=1500 HTTPHeaders=\[]`))
 				Expect(err).NotTo(HaveOccurred())
 				Eventually(sess).Should(Exit(0))
 
 				sess, err = cmd.Start("deis healthchecks:list -a %s", &user, app.Name)
 				Eventually(sess).Should(Say("=== %s Healthchecks", app.Name))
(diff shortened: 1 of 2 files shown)


bacongobbler:
I think instead of changing the existing tests, we should just add one test that confirms setting a healthcheck with a bad port will roll back.

kmala:
@bacongobbler i will write it after i merge this as that test won't pass until https://github.com/deis/controller/pull/1036 is merged.

bacongobbler:
Why not wait until after deis/controller#1036 is merged first so you save yourself from writing 2 separate PRs?

kmala:
https://github.com/deis/controller/pull/1036 won't pass e2e until this is merged.

bacongobbler:
Why is that? I'm probably missing some background context here that's not obvious to me (sorry)

kmala:
this is not adding the e2e tests for the https://github.com/deis/controller/pull/1036 pr...this is fixing the existing e2e tests.

bacongobbler:
The point I'm missing here is *why* this fixes the existing e2e tests. As far as I can tell [e2e has been green for days](https://ci.deis.io/job/controller/).

kmala:
e2e is green for now because the code doesn't have https://github.com/deis/controller/pull/1036 which fails a deploy if the healthcheck fails...current code doesn't do anything if the healthcheck fails and our deploys are failing with https://github.com/deis/controller/issues/989 in e2e but we are not checking it as the command exists 0 status.

bacongobbler:
So the missing context that was never provided was "where did port 1500 and port 8080 come from"?

It turns out that port 1500 is the port number for [`deis/example-dockerfile-http`](https://github.com/deis/workflow-e2e/blob/5bc8cb0f487bc09887e50da7ee2409e22debd134/tests/cmd/builds/commands.go#L18), and port 8080 is for [quay.io/deisci/e2e-private-registry-test](https://github.com/kmala/workflow-e2e/blob/dc9df13d12c82bc5cc7821d633db850e6e9af7d7/tests/registry_test.go#L110). In that case, can you please remove [this comment](https://github.com/deis/workflow-e2e/blob/master/tests/healthcheck_test.go#L69) and add a new comment stating why port 1500 and 8080 are significant?

kmala:
they are the actual ports of the images we are deploying.
i can change that test case.

tests/healthcheck_test.go
```diff
@@ -65,34 +65,33 @@ var _ = Describe("deis healthchecks", func() {
 				Eventually(sess).Should(Exit(0))
 			})
 
-			// disable httpGet liveness checks as it won't pass for every app
-			// TODO(bacongobbler): somehow determine *what* port we need to set the liveness check for this app
-			XSpecify("that user can set a httpGet liveness healthcheck", func() {
-				sess, err := cmd.Start("deis healthchecks:set liveness httpGet -a %s 80", &user, app.Name)
+			// 1500 is the port of th app we are deploying deis/example-dockerfile-http
```
bacongobbler: typo: the*

kmala: changed

tests/healthcheck_test.go
```diff
@@ -65,34 +65,33 @@ var _ = Describe("deis healthchecks", func() {
 				Eventually(sess).Should(Exit(0))
 			})
 
-			// disable httpGet liveness checks as it won't pass for every app
-			// TODO(bacongobbler): somehow determine *what* port we need to set the liveness check for this app
-			XSpecify("that user can set a httpGet liveness healthcheck", func() {
-				sess, err := cmd.Start("deis healthchecks:set liveness httpGet -a %s 80", &user, app.Name)
+			// 1500 is port of th app we are deploying deis/example-dockerfile-http
```
bacongobbler: the typo is still present:

> port of th app

bacongobbler: can you please change this back to "the port of the app"?

PR #324, opened 2016-09-07
fix(docker-test-integration.sh): check workflow-cli-release bucket, too.

vdice:
Credit to @jchauncey for finding the issue.  (When CLI_VERSION is set to a release version, we were unable to find/dl the binary as we were not checking the release bucket.)


vdice:
quick update: changed `both` to `multiple` in comment above

Issue #329, opened 2016-10-05
git_push_test.go - deis run spec flakiness


vdice:
CI is seeing the following two specs fail intermittently -- but consistently enough to fail ~1 out of 5 test runs.

>Deis Workflow.git push deis master with an existing user who has added their public key and who has a local git repo containing buildpack source code and has run `deis apps:create` from within that repo and can execute deis run successfully

>Deis Workflow.git push deis master with an existing user who has added their public key and who has a local git repo containing dockerfile source code and has run `deis apps:create` from within that repo and can execute deis run successfully

The first `buildpack` variant seems to be even more prevalent.  Here are examples of jobs failed by either/both of these two:

https://ci.deis.io/job/workflow-test-pr/7677
https://ci.deis.io/job/workflow-test-pr/7677/artifact/7677/logs/

https://ci.deis.io/job/workflow-test/3241/
https://ci.deis.io/job/workflow-test/3241/artifact/3241/logs/

Example output:
```
10:41:04 ------------------------------
10:44:14 $ deis auth:register http://deis.k8s.local --username=test-481972137 --password=asdf1234 --email=test-481972137@deis.io
10:44:14 Registered test-481972137
10:44:14 Logged in as test-481972137
10:44:14 Configuration file written to /tmp/deis-workflow-home399849476/.deis/test-481972137.json
10:44:14 $ deis keys:add /tmp/deis-workflow-home399849476/.ssh/deiskey-457.pub
10:44:14 Uploading deiskey-457.pub to deis... done
10:44:14 $ deis apps:create test-231794756 
10:44:14 Creating Application... ...done, created test-231794756
10:44:14 Git remote deis successfully created for app test-231794756.
10:44:14 $ GIT_SSH=/tmp/deis-workflow-home399849476/.ssh/git-ssh GIT_KEY=/tmp/deis-workflow-home399849476/.ssh/deiskey-457 git push deis master
10:44:14 Warning: Permanently added '[deis-builder.k8s.local]:2222,[10.15.254.209]:2222' (ECDSA) to the list of known hosts.
10:44:14 remote: ---> Starting build... but first, coffee!        
10:44:14 remote: -----> Restoring cache...        
10:44:14 remote:        No cache file found. If this is the first deploy, it will be created now.        
10:44:14 remote: -----> Go app detected        
10:44:14 remote: -----> Checking Godeps/Godeps.json file.        
10:44:14 remote: -----> Installing go1.7.1... done        
[...]

PR #330, opened 2016-10-06
fix(tests/git_push_test.go): bump deis run timeout

vdice:
Ref https://github.com/deis/workflow-e2e/issues/329
Ref https://github.com/deis/controller/issues/1106


ultimateboy:
I added some additional `deis run` commands in the tests here: https://github.com/deis/workflow-e2e/pull/328/files

Should I increase the timeout on those as well?

vdice:
@ultimateboy at this point in time, yes, I'd increase those as well.  I'll add a TODO to the controller ticket to decrease, if possible, when/if the issue gets address.

PR #334, opened 2016-10-22
chore(apps_test): Fix up logger tests for new log format

jchauncey:
this is for https://github.com/deis/logger/pull/123

### DIFF
--- a/tests/apps_test.go
+++ b/tests/apps_test.go
@@ -150,7 +150,6 @@ var _ = Describe("deis apps", func() {
 
 			uuidRegExp := `[0-9a-f]{8}-([0-9a-f]{4}-){3}[0-9a-f]{12}`
 			procsRegexp := `(%s-[\w-]+) up \(v\d+\)`
-
 			var app model.App
 
 			BeforeEach(func() {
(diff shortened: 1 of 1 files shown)


tests/apps_test.go
```diff
@@ -177,11 +178,24 @@ var _ = Describe("deis apps", func() {
 				Eventually(sess).Should(Exit(0))
 			})
 
+			//: INFO config %s updated
```
vdice: Are all the commented lines being kept when/if getting the timestamp regex to work is revisited?  If so, could an issue tracking this be created and a corresponding comment be added here?

vdice:
@jchauncey if you add `requires logger#123` to the actual git commit message (and force push here), the e2e run should pick it up (you can also list `requires fluentd#61` to grab that as well) (see also https://deis.com/docs/workflow/contributing/submitting-a-pull-request/#cross-repo-commits)

tests/apps_test.go
```diff
@@ -150,6 +150,7 @@ var _ = Describe("deis apps", func() {
 
 			uuidRegExp := `[0-9a-f]{8}-([0-9a-f]{4}-){3}[0-9a-f]{12}`
 			procsRegexp := `(%s-[\w-]+) up \(v\d+\)`
+			// logFormatRegExp := `(\d\d\d\d-\d\d-\d\d)T(\d\d:\d\d:\d\d)((\+|-)\d\d:\d\d)\s(deis\[controller\]:)\s`
```
vdice: Shall we remove this as well as it isn't currently being used?

tests/apps_test.go
```diff
@@ -180,8 +181,11 @@ var _ = Describe("deis apps", func() {
 			Specify("that user can retrieve logs for that app", func() {
 				sess, err := cmd.Start("deis logs -a %s", &user, app.Name)
 				Eventually(sess).Should(SatisfyAll(
-					Say("INFO \\[%s\\]: %s created initial release", app.Name, user.Username),
-					Say("INFO \\[%s\\]: domain %s added", app.Name, app.Name)))
+					Say(`(.+) (deis\[controller\]: INFO config test\-.* updated)`),
+					Say(`(.*) (deis\[controller\]: INFO test\-.* created initial release)`),
+					Say(`(.*) (deis\[controller\]: INFO appsettings test\-.* updated)`),
+					Say(`(.*) (deis\[controller\]: INFO domain test\-.* added)`),
+					Say(`(.*) (deis\[controller\]: INFO build test\-.* created)`)))
```
vdice: Can we add a check for the formatting of app-specific logging, e.g.
```
2016-10-24T19:29:15+00:00 kabuki-fireball[cmd.v2.athvd]: 2016/10/24 19:29:15 10.128.2.21:57488 GET /
```
Perhaps by curling the test app endpoint and then checking logs (minimally for `<appname>[<type>.<release>.<process_id>`)?

jchauncey: alright so this wont work with the app we deploy. It does not generate any log output so for now I wont worry about adding this

tests/apps_test.go
jchauncey: i removed this btw

PR #343, opened 2016-12-08
fix(chart): s/ginko/ginkgo

vdice:


mboersma:
I misspell "ginkgo" every other time I type it.

PR #344, opened 2016-12-15
feat(auth_test.go): add interactive register spec

vdice:
The caveat here is the _semi-interactive_ compromise.  I could not get interactive password entry (and re-entry for confirmation) to work -- the cli always claimed the passwords did not match.  After an initial dive/experimentation into the `gexpect` library code we use here, I determined quite a bit of effort would be needed to shape up the code to begin to debug... 

Therefore, I believe our bases are still covered (in checking behavior after an interactive register -- albiet only with username and email here).  I can file a follow-up issue to revisit the aforementioned conundrum, if that is deemed necessary.

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,5 +1,5 @@
-hash: 3e43564ccccf3c46aa70fb8028d93ce53e90cfc9db09d979d5c46ddf0734d09e
-updated: 2016-08-26T10:25:33.355599586-06:00
+hash: d4df07af09ac276ba107aff25a4019af3f670a70d5b6b453bddf6828b0a5c376
+updated: 2016-12-14T14:29:18.559567025-07:00
 imports:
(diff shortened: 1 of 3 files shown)


bacongobbler:
I worked with @vdice on this particular problem and can confirm it was hell to debug. This is a great alternative. LGTM

Joshua-Anderson:
Nice! Deis V1 used to use gexpect with the python CLI, but when porting over to golang I had similar problems and wasn't able to track them down.

bacongobbler:
> but when porting over to golang I had similar problems and wasn't able to track them down.

The mystery continues...

mboersma:
This is fine since it constitutes a test that would have caught the regression fixed by deis/workflow-cli#278. We can figure out the `gexpect` dain bramage later.

PR #348, opened 2017-02-03
chore(Dockerfile): bump go-dev to v0.22.0

vdice:
Closes https://github.com/deis/docker-go-dev/issues/64


vdice:
oops, I goofed!  I thought I had properly tested this locally by building and then installing the e2e chart with the image updated _however_ I had conveniently forgotten to actually push the local build... so the run was using an older/0.19.0-based image :(  Looks like core issue remains...

Issue #350, opened 2017-03-07
test(s) for default registration mode of admin_only


vdice:
As of Workflow v2.12.0, the default registration mode has changed to `admin_only`.  The temporary workaround for e2e is to install the chart with `controller.registration_mode=enabled` as this is how the current suite of tests expects the cluster to be set up.  

We will now need tests and/or a new suite to exercise the new default of `admin_only`.

vdice:
At high level, this is the following strategy we're going to pursue for this issue:

- leave one failing non-admin `auth:register` spec in the tests, expecting it to fail
- rewrite the register helper method to assume it's an admin registering the user (which should fix the rest of the tests)
- revert the `--set controller.registration_mode=enabled` e2e-runner flag so we are testing the new default

PR #351, opened 2017-03-14
ref(tests): update tests for new registration mode default

vdice:
Ref https://github.com/deis/workflow-e2e/issues/350

Closes https://github.com/deis/workflow-e2e/issues/350

### DIFF
--- a/tests/cmd/auth/commands.go
+++ b/tests/cmd/auth/commands.go
@@ -2,11 +2,13 @@ package auth
 
 import (
 	"fmt"
+	"os"
 
 	"github.com/deis/workflow-e2e/tests/cmd"
 	"github.com/deis/workflow-e2e/tests/model"
 	"github.com/deis/workflow-e2e/tests/settings"
 
+	gexpect "github.com/ThomasRooney/gexpect"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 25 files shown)


tests/cmd/auth/commands.go
```diff
@@ -64,6 +82,13 @@ func Login(user model.User) {
 	Eventually(sess).Should(Say(fmt.Sprintf("Logged in as %s\n", user.Username)))
 }
 
+// RegisterAndLogin registers a user using a randomized username and then logs in as the registered user.
+func RegisterAndLogin() model.User {
```
vdice: @mboersma what do you think about the new use of `RegisterAndLogin`?  I'd also be fine with just keeping it all in `Register` _but_ the code does need to change from registering w/ `--login=false` and then logging in a user separately b/c:
```
// The 'login' flag is set to false during registration because otherwise the newly registered user's configuration
// will be automatically written to the file/profile used to register -- which in this case is the admin user's.
// Since we don't want to overwrite the admin user's configuration, we log the new user in separately.
```

mboersma: Looks like a good solution. :+1:

PR #362, opened 2017-05-30
feat(chart): use upstream stable/spotify-docker-gc chart

vdice:
TODO:
 - [x] test published chart, which now should fetch and bundle in `stable/spotify-docker-gc` (once ci runs, should be available from the `workflow-e2e-pr` charts repo; will update)

**edit**: chart artifact version is `v2.11.2-20170530233955-sha.f9c6a4a` (sub-chart fetched/bundled as expected; see https://ci.deis.io/job/workflow-e2e-chart-publish/1/console)


vdice:
Jenkins, test this please

vdice:
Bumping this up in priority: starting to realize many ci runs have been failing due to the former workflow-e2e chart's use of the spotify/docker-gc daemonsets (they constantly are in `CrashLoopBackoff` and so when we install the chart in ci with `--wait`, the chart frequently fails to meet this condition).

vdice:
e2e actually passed https://ci.deis.io/job/workflow-chart-e2e/323/console  -- just happened to bump up against the total job timeout of 30 mins...

PR #363, opened 2017-05-30
fix(tests/apps_test.go): do not log lengthy byte output on test failure

### LINKED ISSUE #360: apps:run test with lengthy output -- don't log actual output on failure
When/if this [apps:run test](https://github.com/deis/workflow-e2e/blob/b65bed77e653a083e7a32554709d19a7df9522e1/tests/apps_test.go#L218-L223) fails, the `Expect` statement proceeds to log the entire command's output, corrupting the user's terminal and crowding out logs with byte code.

We should re-jigger the test, perhaps checking first if the byte array does not match expected (outside of `Expect`) and then, if not, logging some standard error/failure message accordingly.

vdice:
Closes #360


vdice:
Darn; it seems ginkgo logs output if/when test fails; no matter how it is failed.  Need to find out how to disable this for a particular test...

vdice:
fix/solution for https://github.com/deis/workflow-e2e/issues/360 looks like it needs a diff approach; closing for now.

PR #364, opened 2017-08-02
fix(apps_test.go): temporarily disable spec that fails on GKE

mboersma:


mboersma:
I've gone back and run releases v2.16.0 and v2.15.0 in Jenkins and they're both breaking on this test, which suggests it may be environment-related since those bits didn't change and passed CI previously. I hate to skip good tests but I haven't been able to isolate the core issue and I'm hoping to get a Workflow release done.

